PR #3, opened 2021-08-10
Encryption and compression support

AnatolyPopov:
This PR adds compression and encryption based on RSA key pair provided using configuration arguments.

For now the rotation of the keys is not supported.


s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3RemoteStorageManagerConfig.java
```diff
@@ -45,8 +46,29 @@
...
+    public static final String PRIVATE_KEY = REMOTE_STORAGE_MANAGER_CONFIG_PREFIX + "s3.private_key_pem";
+    private static final String PRIVATE_KEY_DOC = "Private key for storage encryption";
+
+    public static final String IO_BUFFER_SIZE = REMOTE_STORAGE_MANAGER_CONFIG_PREFIX + "s3.io.buffer.size";
+    private static final int IO_BUFFER_SIZE_DEFAULT = 8_192;
+    private static final String IO_BUFFER_SIZE_DOC = "Buffer size for uploading";
+
+    public static final String S3_STORAGE_UPLOAD_PART_SIZE =
+            REMOTE_STORAGE_MANAGER_CONFIG_PREFIX + "s3.upload.part.size";
+    private static final int S3_STORAGE_UPLOAD_PART_SIZE_DEFAULT = 1 << 19;
```
willyborankin, line 61: I think better to use `524288`

build.gradle
```diff
@@ -14,7 +14,16 @@
  * limitations under the License.
  */
 
-allprojects {
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'info.solidsoft.gradle.pitest:gradle-pitest-plugin:1.6.0'
```
willyborankin, line 22: why do we need this plugin?

build.gradle
```diff
@@ -53,25 +71,27 @@ allprojects {
     }
 
     ext {
-        testcontainersVersion = "1.15.3"
-        slf4jVersion = "1.7.30"
+        testcontainersVersion = "1.16.0"
+        zstdVersion = "1.5.0-4"
```
willyborankin, line 75: let's keep the same version as in kafka.

build.gradle
```diff
@@ -53,25 +71,27 @@ allprojects {
...
     dependencies {
-        implementation name: 'kafka-storage-api-3.0.0-SNAPSHOT'
-        implementation name: 'kafka-clients-3.0.0-SNAPSHOT'
-        implementation 'org.slf4j:slf4j-api:1.7.30'
-        testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.7.2'
-        testImplementation 'org.junit.jupiter:junit-jupiter-api:5.7.2'
+        implementation name: 'kafka-storage-api-3.1.0-SNAPSHOT'
+        implementation name: 'kafka-clients-3.1.0-SNAPSHOT'
+        implementation 'org.slf4j:slf4j-api:1.7.32'
+        implementation "com.github.luben:zstd-jni:$zstdVersion"
```
willyborankin, line 83: this dependency must be transient. kafka uses the same library

commons/build.gradle
```diff
@@ -0,0 +1,8 @@
+dependencies {
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.12.4'
```
willyborankin, line 2: same here. plz check tha Kafka does not have such deps

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-all.zip
```
willyborankin, line 3: let's split up the bump up of Gradle version and this solution.

s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3ClientWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+        try {
+            final GetObjectRequest getObjectRequest = new GetObjectRequest(config.s3BucketName(), key);
+            final S3Object s3Object = s3Client.getObject(getObjectRequest);
+            final BoundedInputStream boundedInputStream =
+                    new BoundedInputStream(cryptoIOProvider.decryptAndDecompress(s3Object.getObjectContent()), end);
+            inputStreams.add(boundedInputStream);
+        } catch (final IOException e) {
+            throw new RemoteStorageException(String.format("Error fetching log segment data from %s", key), e);
+        }
+        return new SequenceInputStream(Collections.enumeration(inputStreams));
```
willyborankin, line 111: What I do not like about this solution is that you store all streams/bytes in the memory. Is it possible to use it differently e.g. just create an iterator that will build `InputStream` on demand?
And yet another question `BoundedInpytStream` was introduced for `ServletInputStreams` since they could be blocked due to `Content-Length` why do we use it? What is the diff compare to `ByteArrayInputStream`?

willyborankin (changes requested):
Could plz check all dependencies libs that tey do not in conflict with Kafka ones?
1. ztsd
2. commons-io
3. jackson

s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3ClientWrapper.java
AnatolyPopov: As far as I understand the data will be fetched from S3 only during the reading of the result stream since in the javadoc of `S3Object#getObjectContent` specified that stream will be created on demand(AFAIU):

> Note: The method is a simple getter and does not actually create a stream. If you retrieve an S3Object, you should close this input stream as soon as possible, because the object contents aren't buffered in memory and stream directly from Amazon S3. 

As for `BoundedInputStream` it seems to be suitable because we can limit the content up to specified length. That's what we actually need - read up to end position. And not eagerly like it would be with `ByteArrayInputStream`because it's constructor accepts only byte arrays but on demand later. But maybe I'm missing something.

s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3RemoteStorageManagerConfig.java
AnatolyPopov: Fixed

build.gradle
AnatolyPopov: Fixed

commons/build.gradle
AnatolyPopov: Yes, it has. Changed to the version from Kafka.

gradle/wrapper/gradle-wrapper.properties
AnatolyPopov: Ok

build.gradle
AnatolyPopov: Done.

build.gradle
AnatolyPopov: I'm using it for mutation testing locally. Removed for now.

s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3ClientWrapper.java
willyborankin: Yes, you are right to use `ByteArrayInputStream` we need to read everything into memory. I still dout that we need `BoundedInputStream`. Lets see how it will behave

build.gradle
```diff
@@ -53,25 +61,25 @@ allprojects {
...
+        implementation name: 'kafka-clients-3.1.0-SNAPSHOT'
+        implementation 'org.slf4j:slf4j-api:1.7.32'
+        implementation 'org.bouncycastle:bcprov-jdk15on:1.69'
+        testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
+        testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
         testImplementation 'org.apache.kafka:kafka_2.12:2.8.0'
         testImplementation "org.testcontainers:junit-jupiter:$testcontainersVersion"
         testImplementation "org.testcontainers:kafka:$testcontainersVersion" // this is not Kafka version
-        testImplementation "org.assertj:assertj-core:3.19.0"
+        testImplementation 'org.assertj:assertj-core:3.20.2'
```
willyborankin, line 78: better to use `"` instead of `'`

willyborankin (dismissed):
Except for one small note LGTM.

build.gradle
AnatolyPopov: Fixed

PR #131, opened 2023-03-21
Add chunk indices

ivanyu:


commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkIndex.java
```diff
@@ -0,0 +1,14 @@
+package io.aiven.kafka.tiered.storage.commons.chunkindex;
```
AnatolyPopov, line 1: Copyrights?

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkIndexWithoutCompressionTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class ChunkIndexWithoutCompressionTest {
+    @Test
+    void incorrectConstructorParams() {
+        assertThatThrownBy(() -> new ChunkIndexForFilesWithoutCompression(-1, 100, 100),
+            "Original chunk size cannot be negative, -1 given")
```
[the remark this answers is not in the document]
ivanyu: Great, thanks! Fixed

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
```diff
@@ -0,0 +1,165 @@
...
+        this.chunksAdded += 1;
+    }
+
+    protected void addChunk0(final int transformedChunkSize) {}
+
+    @Override
+    public final void finish(final int finalTransformedChunkSize) {
+        if (finished) {
+            throw new IllegalStateException("Cannot finish already finished index");
+        }
```
ivanyu, line 62: This whole mechanics around finishing an index asks for a separate builder class, but probably a bit later

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkIndexBuilderCommonTest.java
```diff
@@ -0,0 +1,128 @@
...
+        assertThatThrownBy(() -> index.findChunkForOriginalOffset(-1))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Offset must be non-negative, -1 given");
+    }
+
+    @Test
+    void findForEmptyFile() {
+        final ChunkIndex index = emptyFileChunkIndexBuilder.finish(0);
+
+        assertThat(index.findChunkForOriginalOffset(0)).isEqualTo(null);
```
AnatolyPopov, line 109: Nit:
```suggestion
        assertThat(index.findChunkForOriginalOffset(0)).isNull();
```

And below also

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ChunkTest {
+    @Test
+    void identical() {
+        final var c1 = new Chunk(0, 20, 30, 100, 120);
+        final var c2 = new Chunk(0, 20, 30, 100, 120);
+        assertThat(c1.equals(c2)).isTrue();
+        assertThat(c2.equals(c1)).isTrue();
+        assertThat(c1.hashCode()).isEqualTo(c2.hashCode());
```
AnatolyPopov, line 30:
```suggestion
        assertThat(c1).hasSameHashCodeAs(c2);
```
And there is also `doesNotHaveSameHashCodeAs` for other places

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
```diff
@@ -0,0 +1,67 @@
...
+    // it should not be persisted.
+    private final List<Chunk> chunks;
+
+    public VariableSizeChunkIndex(final int originalChunkSize,
+                                  final int originalFileSize,
+                                  final List<Integer> transformedChunks) {
+        super(originalChunkSize, originalFileSize, transformedChunks.get(transformedChunks.size() - 1));
+        this.transformedChunks = List.copyOf(Objects.requireNonNull(transformedChunks));
+
+        chunks = materializeChunks();
```
AnatolyPopov, line 51: From Effective Java:
> There are a few more restrictions that a class must obey to allow inheritance. Constructors must not invoke overridable methods, directly or indirectly. If you violate this rule, program failure will result. The superclass constructor runs before the subclass constructor, so the overriding method in the subclass will be invoked before the subclass constructor has run. If the overriding method depends on any initialization performed by the subclass constructor, the method will not behave as expected.

TLDR: we should not call overridable methods from constructors I think.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
```diff
@@ -0,0 +1,67 @@
...
+
+    // This only a materialization for convenience and performance,
+    // it should not be persisted.
+    private final List<Chunk> chunks;
+
+    public VariableSizeChunkIndex(final int originalChunkSize,
+                                  final int originalFileSize,
+                                  final List<Integer> transformedChunks) {
+        super(originalChunkSize, originalFileSize, transformedChunks.get(transformedChunks.size() - 1));
+        this.transformedChunks = List.copyOf(Objects.requireNonNull(transformedChunks));
```
AnatolyPopov, line 49: Something strange seems to be going on here, if we get to this line `transformedChunks` should not be null already I guess.

AnatolyPopov, line 49: Also I think builder is not reusable, right? So it perhaps does not make sense to make a copy here since the constructor is called only in builder's `finish0`

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
```diff
@@ -0,0 +1,67 @@
...
+
+        chunks = materializeChunks();
+    }
+
+    public List<Chunk> chunks() {
+        return chunks;
```
AnatolyPopov, line 55: Instead of copying the list in the constructor I guess we would want to return a copy here to keep this list unmodifiable.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/FixedSizeChunkIndex.java
```diff
@@ -0,0 +1,70 @@
...
+    public FixedSizeChunkIndex(final int originalChunkSize,
+                               final int originalFileSize,
+                               final int transformedChunkSize,
+                               final int finalTransformedChunkSize) {
+        super(originalChunkSize, originalFileSize, finalTransformedChunkSize);
+
+        checkSize(transformedChunkSize, "Transformed chunk size");
+        this.transformedChunkSize = transformedChunkSize;
+
+        chunks = materializeChunks();
```
AnatolyPopov, line 53: Same as for `VariableSizeChunkIndex` and overridable method calls in constructor.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/FixedSizeChunkIndex.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public List<Chunk> chunks() {
+        return chunks;
```
AnatolyPopov, line 57: Same as in `VariableSizeChunkIndex` I would return a copy here.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: I think this applies to the case where the `AbstractChunkIndex` constructor calls `materializeChunks` and its subsequent calls. However when `FixedSizeChunkIndex` and `VariableSizeChunkIndex` does this, it should be fine as all the necessary state is initialized. I can't find a better way to not repeat the `materializeChunks` code in both classes.

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ChunkTest {
+    @Test
+    void identical() {
+        final var c1 = new Chunk(0, 20, 30, 100, 120);
+        final var c2 = new Chunk(0, 20, 30, 100, 120);
+        assertThat(c1.equals(c2)).isTrue();
```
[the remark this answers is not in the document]
ivanyu: I kinda did this intentionally to show I'm checking the result of `equals` as a thing itself

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndexBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tiered.storage.commons.chunkindex;
+
+public abstract class AbstractChunkIndexBuilder implements ChunkIndexBuilder {
+    protected final int originalChunkSize;
+    protected final int originalFileSize;
+
+    protected int chunksAdded = 0;
```
AnatolyPopov, line 23: Does this need to be protected? It's only used in this class.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndexBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+        if (remainOfOriginalFileSize() <= originalChunkSize) {
+            throw new IllegalStateException("This must be final chunk. Call `finish` instead.");
+        }
+
+        addChunk0(transformedChunkSize);
+
+        this.chunksAdded += 1;
+    }
+
+    protected void addChunk0(final int transformedChunkSize) {
```
AnatolyPopov, line 52: Should this one be abstract?

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndexBuilder.java
ivanyu: Nah, same for `finished`. Will make `private`.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: I mean, `materializeChunks` produce already an unmodifiable list

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: That's true, the builder is one-time. however I kept the check for basic hygiene. At some point soon we'll be deserializing those indices and also directly instantiating them in tests.

ivanyu: That's true, the builder is one-time. However, the same reason as for keeping `Objects.requireNonNull`

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndexBuilder.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package io.aiven.kafka.tiered.storage.commons.chunkindex;
+
+public abstract class AbstractChunkIndexBuilder implements ChunkIndexBuilder {
+    protected final int originalChunkSize;
+    protected final int originalFileSize;
+
+    protected int chunksAdded = 0;
+    protected boolean finished = false;
```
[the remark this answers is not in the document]
ivanyu: I like `finished` more because "closing" doesn't seem to make much sense on a builder

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkTest.java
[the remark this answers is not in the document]
AnatolyPopov: The difference is in the output of test failure. `isEqualTo` prints the actual objects side by side so it's easier to spot why the equality check is failing. Regular `isTrue` however will print only that test expected true but the result was false and imo this brings a little less value. But anyways this is nitpicking and you can leave it as it is if you want.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
AnatolyPopov: Maybe make `materializeChunks` final then? I understand your point and do agree but at the same time I can imagine that somebody might want to patch `FixedSizeChunkIndex` or `VariableSizeChunkIndex` for their use case using inheritance since the classes are not final also. So this is not a problem by itself right now but a potential problem in the future that will be hard to spot if there will be the case.

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/chunkindex/ChunkTest.java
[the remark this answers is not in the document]
ivanyu: Ah nvm, did this change anyway

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
AnatolyPopov: As you wish but I do not see any value neither in copying it here nor making it unmodifiable since it's a private field.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
AnatolyPopov: Oh, sorry I was thinking about different field when mentioned constructor.
Regarding `materializeChunks` looks like I'm missing some context here. Is the idea here also to cache the produced list as a field in the instance of this object and that is the reason why we are calling `materializeChunks` from constructor and not from here directly? And moreover this *Index classes supposed to be immutable, correct?(This I guess should be mentioned somewhere for clarity)

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: Well, you're right. I removed the copying

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: > Is the idea here also to cache the produced list as a field in the instance of this object and that is the reason why we are calling `materializeChunks` from constructor and not from here directly?

Yeah, the idea is to pre-calculate the chunks list and cache it. It makes sense because the chunk list is inevitable to be used and potentially multiple times. It's done in the constructor, because it's automatically thread-safe this way (no need to have locks/`volatile`).

> And moreover this *Index classes supposed to be immutable, correct?(This I guess should be mentioned somewhere for clarity)

Yes, you're right. I added this clarification.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
```diff
@@ -0,0 +1,123 @@
...
+        checkSize(originalFileSize, "Original file size");
+        this.originalFileSize = originalFileSize;
+
+        checkSize(finalTransformedChunkSize, "Final transformed chunk size");
+        this.finalTransformedChunkSize = finalTransformedChunkSize;
+    }
+
+    // Override with care, this method is indirectly called from a constructor.
+    protected final List<Chunk> materializeChunks() {
+        final List<Chunk> chunks = new ArrayList<>(chunkCount());
```
AnatolyPopov, line 42: `materializeChunks` is now final but `chunkCount` is not and I guess there is exactly the same issue with a call from constructor here.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
```diff
@@ -0,0 +1,123 @@
...
+        int originalPosition = 0;
+        int transformedPosition = 0;
+        if (chunkCount() == 0) {
+            chunks.add(new Chunk(0, 0, 0, 0, 0));
+        } else {
+            for (int chunkI = 0; chunkI < chunkCount(); chunkI++) {
+                final int originalSize = originalChunkSize(chunkI);
+                final int transformedSize = transformedChunkSize(chunkI);
```
AnatolyPopov, line 50: Same thing as above.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
AnatolyPopov: > >Is the idea here also to cache the produced list as a field in the instance of this object and that is the reason why we are calling materializeChunks from constructor and not from here directly?
>
> Yeah, the idea is to pre-calculate the chunks list and cache it. It makes sense because the chunk list is inevitable to be used and potentially multiple times. It's done in the constructor, because it's automatically thread-safe this way (no need to have locks/volatile).

I'm starting to hesitate that it's a right approach taking into account all potential initialization issues. Probably we could use a similar approach but move this `materializeChunks` out of this hierarchy to some kind of wrapper maybe..

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
AnatolyPopov: And AFAIU an option probably could be to make `chunkCount` a final field initialized in constructor instead since all the info is available in constructor to calculate it.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
ivanyu: What exactly?

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
ivanyu: That's a good point, I made `chunkCount` a final field

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
ivanyu: Extracting it in any form will not work as the algorithm inextricably depends on the object's state: in one way or another we'll have to pass closures on the state. The only way to make it "right" is to make two copies of `transformedChunkSize` and `originalChunkSize`, one to be called in the constructor and the other in other places, which doesn't make sense.

The book's advice is generally right, but it in this particular case there is no issue. The present code execution is correct and we left the warnings for the posterity in case the classes are to be modified or extended. I don't see how realistically it can fail.

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/AbstractChunkIndex.java
AnatolyPopov: Was thinking about `transformedChunkSize` being overridable, but nevermind

commons/src/main/java/io/aiven/kafka/tiered/storage/commons/chunkindex/VariableSizeChunkIndex.java
AnatolyPopov: Revisited it once again and with additions of finals seems to be fine now.

AnatolyPopov (approved):
LGTM

PR #139, opened 2023-03-30
Add chunk transformations

ivanyu:
This commit introduces chunk transformations, that are the foundation of the chunking itself (both for upload and download) and also of the optional encryption and compression.


ivanyu:
A note for the reviewer:
There are two independent entry points to the review: `TransformChunkEnumeration` (with its descendants and tests) and `DetransformChunkEnumeration` (with its descendants and tests).  They overlap only in `TransformsEndToEndTest`.

build.gradle
```diff
@@ -79,10 +79,13 @@ subprojects {
         testImplementation name: "kafka-storage-api-3.1.0-SNAPSHOT"
         testImplementation name: "kafka-clients-3.1.0-SNAPSHOT"
         testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
+        testImplementation "org.junit.jupiter:junit-jupiter-params:$junitVersion"
         testImplementation "org.apache.kafka:kafka_2.12:2.8.0"
         testImplementation "org.testcontainers:junit-jupiter:$testcontainersVersion"
         testImplementation "org.testcontainers:kafka:$testcontainersVersion" // this is not Kafka version
         testImplementation "org.assertj:assertj-core:3.24.2"
+        testImplementation 'org.mockito:mockito-core:5.1.1'
+        testImplementation 'org.mockito:mockito-junit-jupiter:5.1.1'
```
AnatolyPopov, line 88: Should we have a variable for version?

ivanyu: Yes, fixed

commons/build.gradle
```diff
@@ -14,10 +14,15 @@
...
+}
+
 dependencies {
     // The client of a library should provide a runtime dependency. For now taking the version from kafka-storage-api
     compileOnly 'com.fasterxml.jackson.core:jackson-databind:2.14.2'
     // The client of a library should provide a runtime dependency.
     // For now taking the version from kafka-clients
-    compileOnly "com.github.luben:zstd-jni:1.5.4-2"
+    compileOnly group: "com.github.luben", name: "zstd-jni", version: zstdVersion
+    testImplementation group: "com.github.luben", name: "zstd-jni", version: zstdVersion
```
AnatolyPopov, line 27: Oh, I missed this in another and merged already but should we preserve the style as it was? e.g.
`testImplementation "com.github.luben:zstd-jni:$zstdVersion"

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/transform/TransformsEndToEndTest.java
```diff
@@ -0,0 +1,107 @@
...
+        try (final var sis = new SequenceInputStream(transformFinisher)) {
+            uploadedData = sis.readAllBytes();
+            chunkIndex = transformFinisher.chunkIndex();
+        }
+
+        // Detransform.
+        DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(
+            new ByteArrayInputStream(uploadedData), chunkIndex.chunks());
+        if (encryption) {
+            assert secretKey != null;
```
AnatolyPopov, line 94: Should this be `assertThat(secretKey).isNotNull()` with a proper message explaining the failure?

ivanyu: This is not really a part of the test, just a sanity check for the test code itself. But yeah, we can do this, why not

AnatolyPopov: I understand that this is a sanity check but anyways `assert` I believe should not be used in tests and also can be disabled by JVM flags

commons/src/test/java/io/aiven/kafka/tiered/storage/commons/transform/TransformsEndToEndTest.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private void test(final int chunkSize, final boolean compression, final boolean encryption) throws IOException {
+        // Transform.
+        TransformChunkEnumeration transformEnum = new BaseTransformChunkEnumeration(
+            new ByteArrayInputStream(original), chunkSize);
+        if (compression) {
+            transformEnum = new CompressionChunkEnumeration(transformEnum);
+        }
+        if (encryption) {
```
AnatolyPopov, line 79: I guess we need same check for `secretKey` nullability here, right?

commons/build.gradle
ivanyu: Yeah, let's use one. Updated.

PR #141, opened 2023-03-31
Switch to published version of Kafka deps

ivanyu:


AnatolyPopov (commented):
LGTM but should it be in the main branch?

ivanyu:
We'll bring it to `main` when we merge the whole `new-implementation` there?

PR #144, opened 2023-03-31
chore: introduce docker playground for tiered-storage

jeqo:
Adds a Docker playground to build Kafka from source-code and add plugins for testing purposes.
No guarantee to keep it stable just yet.

The goal is to iterate faster on how users would use the plugins in their environments.

PR #145, opened 2023-03-31
feat: add custom S3 endpoint URL config

jeqo:
While enabling a testing environment (see #144), there is a need to point to localstack or minio (or any s3 compatible endpoint), but plugin currently doesn't expose this configuration.

Depends on #144

### DIFF
--- a/s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3RemoteStorageManager.java
+++ b/s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3RemoteStorageManager.java
@@ -37,7 +37,6 @@ import org.slf4j.LoggerFactory;
  */
 public class S3RemoteStorageManager implements RemoteStorageManager {
     private static final Logger log = LoggerFactory.getLogger(S3RemoteStorageManager.class);
-    private AwsClientBuilder.EndpointConfiguration endpointConfiguration = null;
 
     private S3RemoteStorageManagerConfig config;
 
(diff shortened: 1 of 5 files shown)


ivanyu:
@jeqo wrong target branch, I guess :)

jeqo:
@ivanyu good catch, thanks! updated it to `new-implementation`

jeqo:
(rebased -- commits should look better now, cc @AnatolyPopov )

s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3RemoteStorageManager.java
```diff
@@ -64,7 +64,15 @@ public void configure(final Map<String, ?> configs) {
...
         AmazonS3ClientBuilder s3ClientBuilder = AmazonS3ClientBuilder.standard();
         if (endpointConfiguration == null) {
-            s3ClientBuilder = s3ClientBuilder.withRegion(config.s3Region());
+            if (config.s3EndpointUrl().isEmpty()) {
+                s3ClientBuilder = s3ClientBuilder.withRegion(config.s3Region());
+            } else {
+                final AwsClientBuilder.EndpointConfiguration endpointConfig =
+                    new AwsClientBuilder.EndpointConfiguration(
+                        config.s3EndpointUrl().get(),
+                        config.s3Region().getName());
```
AnatolyPopov, line 73: We should perhaps specify this parameters in tests also and these will allow to simplify all these if statements.

jeqo: Just applied this fixes, let me know how it looks.

PR #156, opened 2023-04-05
feat: add local tiered storage impl to docker options

jeqo:
Adds a new docker compose configuration to use local tiered storage implemented in https://github.com/aiven/kafka/blob/93ae220db12e8294b2cfe8871a2583b53de8b775/storage/src/test/java/org/apache/kafka/server/log/remote/storage/LocalTieredStorage.java, and follows https://github.com/aiven/kafka/blob/3.3-2022-10-06-tiered-storage/storage/src/test/README.md to configure plugin.

This should help us to compare plugin implementations.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -1,0 +1,18 @@
+all:
+
+# Defaults
+t=t1
+segment=10000000 # 10M
+retention=500000000 # 500M
+local_retention=50000000 # 50M
+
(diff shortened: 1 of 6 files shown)


jeqo:
Falling back to draft. We could wait for Local implementation to be on upstream Kafka or on this repo before opening again.

PR #159, opened 2023-04-14
refactor: reorg commons module

jeqo:
changes: 
- rename tieredstorage, 
- move chunk to root, 
- rename index module,

fixes #155

PR #160, opened 2023-04-14
Refactor transform

jeqo:
Depends on #159 

Introduce transform chains to handle all the internal details of transform flows.


jeqo:
Replaced by #217

PR #162, opened 2023-04-16
Add ChunkManager interface

ivanyu:


jeqo (commented):
@ivanyu could you elaborate a bit further how is this interface is going to be used or add a default implementation?
By the signature it seems to be related to get chunks as part of the fetching process, but I'm failing to see how it's going to be used.

ivanyu:
Basically `ChunkManager` is a single point provider of chunks for the download path. It encapsulates downloading itself and caching and exposes only one function for getting a chunk. The code that uses it will need to know what chunks it needs to satisfy its requests.

Here's a draft of the `Enumeration` implementation that can be used to form a fetching `InputStream`:
```java
public class FetchChunkEnumeration implements Enumeration<InputStream> {
    private static final int INPUT_STREAM_FUTURE_WAIT_TIMEOUT_SEC = 30;

    private final ChunkManager chunkManager;
    private final RemoteLogSegmentMetadata remoteLogSegmentMetadata;
    private final SegmentManifest manifest;
    private final int startPosition;

    private final int startChunkId;
    private final int lastChunkId;
    private int curChunkId;

    public FetchChunkEnumeration(final ChunkManager chunkManager,
                                 final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                                 final SegmentManifest manifest,
                                 final int startPosition) {
        this.chunkManager = Objects.requireNonNull(chunkManager, "chunkManager cannot be null");
        this.remoteLogSegmentMetadata = Objects.requireNonNull(remoteLogSegmentMetadata,
            "remoteLogSegmentMetadata cannot be null");
        this.manifest = Objects.requireNonNull(manifest, "manifest cannot be null");

        final Chunk startChunk = manifest.chunkIndex().findChunkForOriginalOffset(startPosition);
        if (startChunk == null) {
            throw new IllegalArgumentException("Invalid start position "
                + startPosition + " in segment " + remoteLogSegmentMetadata);
        }
        this.startPosition = startPosition;

        startChunkId = startChunk.id;
        curChunkId = startChunkId;

        final List<Chunk> chunks = manifest.chunkIndex().chunks();
        lastChunkId = chunks.get(chunks.size() - 1).id;
    }


    @Override
    public boolean hasMoreElements() {
        return curChunkId <= lastChunkId;
    }

    @Override
    public InputStream nextElement() {
        if (!hasMoreElements()) {
            throw new NoSuchElementException();
        }

        final Future<InputStream> chunkF = chunkManager.getChunk(remoteLogSegmentMetadata, manifest, curChunkId);
        final InputStream inputStream;
        try {
            inputStream = chunkF.get(INPUT_STREAM_FUTURE_WAIT_TIMEOUT_SEC, TimeUnit.SECONDS);
        } catch (final InterruptedException | ExecutionException | TimeoutException e) {
        }

        if (curChunkId == startChunkId) {
            final int toSkip = startPosition - manifest.chunkIndex().chunks().get(curChunkId).originalPosition;
            try {
                inputStream.skip(toSkip);
            }
        }

        curChunkId += 1;
        return inputStream;
    }
}
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
+
+import io.aiven.kafka.tieredstorage.commons.manifest.SegmentManifest;
+
+public interface ChunkManager {
+    /**
+     * Gets a chunk of a segment.
+     * @return an {@link InputStream} of the chunk, plain text (i.e. decrypted and decompressed).
+     */
+    Future<InputStream> getChunk(RemoteLogSegmentMetadata remoteLogSegmentMetadata,
```
jeqo, line 31: Could you also elaborate (or add in the doc) why is this async?

AFAIK, the TS framework already wraps remote fetchs on Future:

https://github.com/apache/kafka/blob/a0a17208ad5ac57d35c5bd2c7ec2836ed3cafe66/core/src/main/java/kafka/log/remote/RemoteLogManager.java#L813-L815

jeqo, line 31: Also, the name it's a bit misleading: it says getChunk, but return a Future<InputStream>

jeqo (commented):
@ivanyu thanks for the explanation!
Wouldn't be possible to implement this directly on the FetchChunk transformation proposed? or do we see that it will be used somewhere else?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
ivanyu: It's pretty common to name `getXyz` instead of `getXyzFuture` as the latter clutters the code quite a bit without helping much. I changed the Javadoc a bit to mention that it's a future.

ivanyu:
> Wouldn't be possible to implement this directly on the FetchChunk transformation proposed?

The problem is that `FetchChunkEnumeration` will be created each time a fetching is needed, but there will be one `ChunkManager`, which will be managing all the state and internal mechanics.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
ivanyu: That's a good question. The `Future` approach comes from some early prototyping we did with @AnatolyPopov where it's possible to satisfy multiple identical parallel `getChunk` requests without doing downloading the same segment multiple times. `Future` worked as a very convenient placeholder and synchronization barrier. This could have been done on locks and atomic references, but futures are just more comfortable.

I propose to treat this `ChunkManager` interface as a starting point and change it as we see fit when the real implementation is done, including also removing the `Future` wrapper.

jeqo: Oh, I press "Comment" to quick. I was referring to the inputStream, though I think a similar response would apply.

jeqo (approved):
Oh got it. We can revisit this further when the implementation comes in another PR.
Thanks again @ivanyu

jeqo:
@ivanyu sure, go ahead and merge it 👍🏽

ivanyu:
Done! Feel free to merge.
(Damn, now the RSM tests are failing, but we'll fill this gap soon)

PR #163, opened 2023-04-16
Streamline encryption key and AAD generation

ivanyu:


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/EncryptionKeyProvider.java
```diff
@@ -69,6 +71,15 @@ public SecretKey createKey() {
         return aesKeyGenerator.generateKey();
     }
 
+    public EncryptionKeyAndAAD createKeyAndAAD() {
```
jeqo, line 74: Should we rename the returned class to `SecretKeyAndAAD` instead to keep it aligned with `createKey(): SecretKey`?

jeqo (commented):
LGMT, adding a small comment about naming

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/EncryptionKeyAndAAD.java
```diff
@@ -0,0 +1,51 @@
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final EncryptionKeyAndAAD that = (EncryptionKeyAndAAD) o;
+        return Objects.equals(key, that.key) && Arrays.equals(aad, that.aad);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(key);
```
AnatolyPopov, line 47: I guess this instead
```suggestion
        int result = Objects.hashCode(key);
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/EncryptionKeyProvider.java
ivanyu: > Should we have a test for it also?

Not sure. What to test, that it doesn't return `null`?

ivanyu: > Should we rename the returned class to `SecretKeyAndAAD` instead to keep it aligned with `createKey(): SecretKey`?

Unification of these two terms throughout the code base is indeed needed. I personally would prefer "encryption" over "secret". But this probably goes beyond the scope of this PR. I've made https://github.com/aiven/tiered-storage-for-apache-kafka/issues/166 for this.

PR #164, opened 2023-04-16
Add ObjectKey class

ivanyu:


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
```diff
@@ -0,0 +1,71 @@
...
+package io.aiven.kafka.tieredstorage.commons;
+
+import java.text.NumberFormat;
+import java.util.Objects;
+
+import org.apache.kafka.common.TopicIdPartition;
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentId;
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
+
+public final class ObjectKey {
```
jeqo, line 26:
```suggestion
/**
 * Maps Kafka Segment files to Object paths in the storage backend.
 */
public final class ObjectKey {
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
```diff
@@ -0,0 +1,71 @@
...
+    public enum Suffix {
```
jeqo, line 27:
```suggestion

    /**
     * Supported files and extensions, including: Log, Index types, and Object storage manifest
     *
     * @see org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType
     */
    public enum Suffix {
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static String key(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
+                             final Suffix suffix) {
+        Objects.requireNonNull(remoteLogSegmentMetadata, "remoteLogSegmentMetadata cannot be null");
+        Objects.requireNonNull(suffix, "suffix cannot be null");
+
+        final RemoteLogSegmentId remoteLogSegmentId = remoteLogSegmentMetadata.remoteLogSegmentId();
+        final TopicIdPartition topicIdPartition = remoteLogSegmentId.topicIdPartition();
+
+        // <topic_name>-<topic_uuid>/<partition>/00000000000000000000-<segment_uuid>.<suffix>
```
jeqo, line 50:
```suggestion
    /**
     * @return Object path/key with the following format: <code>
     *     topic_name-topic_uuid/partition/(000+start_offset length=20)-segment_uuid.suffix
     * </code>
     */
    public static String key(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                             final Suffix suffix) {
        Objects.requireNonNull(remoteLogSegmentMetadata, "remoteLogSegmentMetadata cannot be null");
        Objects.requireNonNull(suffix, "suffix cannot be null");

        final RemoteLogSegmentId remoteLogSegmentId = remoteLogSegmentMetadata.remoteLogSegmentId();
        final TopicIdPartition topicIdPartition = remoteLogSegmentId.topicIdPartition();

```

jeqo (approved):
LGTM, added some doc improvements

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
```diff
@@ -0,0 +1,71 @@
...
+        MANIFEST("rsm-manifest");
+
+        public final String value;
+        Suffix(final String value) {
+            this.value = value;
+        }
+    }
+
+    public static String key(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
+                             final Suffix suffix) {
```
AnatolyPopov, line 43: And what about prefix?

ivanyu: I'm thinking if it should be a part of the S3 implementation config and not be exposed to this level...

ivanyu: I still don't know, though. WDYT guys @AnatolyPopov @jeqo? Basically:
- the prefix is an storage implementation-specific setting, like the bucket; or
- the prefix is universal and should be used for all the storage implementations.

ivanyu:
Thanks @jeqo, addressed your comments about the Javadoc.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
AnatolyPopov: I agree, this is a static method and we can provide whatever here when we will use it, so just providing empty fi it's not needed is fine by me. On the other hand if we will not provide it here then we will need to do something on top of this to add a prefix in S3 implementation at least.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Maps Kafka segment files to object paths/keys in the storage backend.
+ */
+public final class ObjectKey {
+    /**
+     * Supported files and extensions, including log, index types, and segment manifest.
+     *
+     * @see org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType
+     */
+    public enum Suffix {
```
AnatolyPopov, line 35: Do we really need this enum? AFAIU we do not use the actual `enum` functionality and to map this suffixes to actual index types we will need to do some mapping so why not just use Kafka's `IndexType` enum like we were doing before and constants for manifest and log files?

jeqo: Agree this is missing. I guess we could have a static method within the enum to do the mapping? 
```java
  public enum Suffix {
        // ...
        static Suffix from(RemoteStorageManager.IndexType indexType) {
            switch (indexType) {
                case LEADER_EPOCH: return LEADER_EPOCH_CHECKPOINT;
                case TRANSACTION: return TXN_INDEX;
                case OFFSET: return OFFSET_INDEX;
                case TIMESTAMP: return TIME_INDEX;
                case PRODUCER_SNAPSHOT: return PRODUCER_SNAPSHOT;
                default: throw new IllegalArgumentException("Unknown index type " + indexType);
            }
        }
  }
```
So, when copying files we could pass Suffix.from(...) when we know the index type, and Suffix.LOG/MANIFEST for the other cases.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
ivanyu: Ok, added the prefix

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
ivanyu: Well, this makes sense. I added this static conversion

jeqo (approved):
LGMT, leaving it to @AnatolyPopov to close the open comments.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
AnatolyPopov: Should the `/` after prefix be added here to avoid keeping in mind that it should be added? The reason is that we sould most probably like to put all the stuff in `sub-directory` in S3 for example

jeqo: Just a though: what if users don't want to make it a directory and just adding a cluster name as prefix to the topic names? e.g. `cluster-1-t1-AAAAA1/...`
We could just document that it's up to them to define whether to make the prefix a sub-directory or not. wdyt?

AnatolyPopov: Perhaps you are right, was thinking from a bit different perspective so nm.

AnatolyPopov (approved):
LGTM

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ObjectKey.java
ivanyu: > Just a though: what if users don't want to make it a directory and just adding a cluster name as prefix to the topic names?

Yeah, that's the idea behind not adding `/`. Since S3 and other object storages don't treat `/` anyhow special, I think it's fair to avoid this on our side too (despite we're doing some `/` ourselves, but that's unavoidable).

> We could just document that it's up to them to define whether to make the prefix a sub-directory or not.

Yeah. After the changes here we'll need to update the RSM config class, so I made a ticket: https://github.com/aiven/tiered-storage-for-apache-kafka/issues/168

ivanyu:
Oh no, fixups remained :)

AnatolyPopov:
yeah, just noticed that I forgot about them :(

ivanyu:
No big deal

PR #165, opened 2023-04-16
Add tests for UniversalRemoteStorageManagerTest

ivanyu:


commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerTest.java
```diff
@@ -0,0 +1,348 @@
...
+    static final RemoteLogSegmentMetadata REMOTE_LOG_METADATA = new RemoteLogSegmentMetadata(
+        REMOTE_SEGMENT_ID, START_OFFSET, 2000L,
+        0, 0, 0, SEGMENT_SIZE, Map.of(0, 0L));
+
+    static final ByteBuffer LEADER_EPOCH_INDEX = ByteBuffer.wrap("leader epoch index".getBytes());
+
+    static final String TARGET_LOG_FILE =
+        "topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.log";
+    static final String TARGET_MANIFEST_FILE =
+        "topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.rsm-manifest";
```
jeqo, line 93: Would it make sense to use ObjectKey #164 to build paths here?

jeqo (approved):
LGMT

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerTest.java
ivanyu: Most of the time I try to avoid this approach if it's not too cumbersome. It's more convenient, but instead of "this test tests a particular expected behavior of the code" it turns into "this test tests that the code is aligned with another piece of code which may be not correct in the first place"

jeqo: Got it, fair enough

PR #169, opened 2023-04-18
chore: separate minio and aws setup

jeqo:
It separates AWS S3 and Minio setups to simplify testing.

PR #171, opened 2023-04-18
[DRAFT] Experimental minimal S3 plugin

jeqo:
Depends on #169 

Implements an S3 RSM directly using the S3 client. 

The purpose of this plugin implementation is to 
- gain experience with the S3 client
- benchmark the features included on `new-implementation` branch (e.g. chunking, encryption, compression) against this to evaluate performance. 

Not meant to be merged at the moment, as main work is going on new-implementation branch.

### DIFF
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -3,6 +3,3 @@
 
 # Demo keys
 *.pem
-
-# Backend volumes
-minio/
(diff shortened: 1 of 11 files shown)


jeqo:
will be partially included into S3 benchmarks, e.g. #273

PR #174, opened 2023-04-20
feat: segment compression checker

### LINKED ISSUE #173: [New implementation] Compress chunks when logs are uncompressed
# What is currently missing?

Compression is applied even when segments are already compressed, leading to ineffective use of CPU time.

# How could this be improved?

We have looked into alternatives to read log segments, and get compression from it. A potential challenges is that compression could change on the same log segment, based on topic configuration.

There are different alternatives:
- Reading the first batch only
- Reading the first and last batches and compare the compression
- Check compression doesn't change on all batches.

As changing compression happens usually at most once per topic, the first alternative seems to be the most efficient.

# Is this a feature you would work on yourself?

* [x] I plan to open a pull request for this feature

jeqo:
Validates if compression is required on a segment or not.

Fix #173

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage.commons;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.FileRecords;
+import org.apache.kafka.common.record.RecordBatch;
+
+/**
+ * Checks if segment are compressed or not.
+ * To be used when segment files are received on archival.
+ */
+public class SegmentCompressionChecker {
+
+    /**
+     * @param file Kafka log segment file
+     * @return true if log segment is compressed, otherwise returns false
+     */
+    public static boolean check(final File file) throws InvalidRecordBatchException {
+        try (final FileRecords records = FileRecords.open(file, false, true, 0, false)) {
+            final RecordBatch batch = fistRecordBatch(records);
+            return batch.compressionType() != CompressionType.NONE;
+        } catch (final IOException | KafkaException e) {
+            throw new InvalidRecordBatchException("Failed to read and validate first batch", e);
+        }
+    }
+
+    private static RecordBatch fistRecordBatch(final FileRecords records) throws InvalidRecordBatchException {
+        final RecordBatch batch = records.firstBatch();
+        if (batch == null) {
+            throw new InvalidRecordBatchException("Record batch is null");
+        }
+        batch.ensureValid();
+        return batch;
+    }
+}
--- a/commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
+++ b/commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage.commons;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.Path;
+
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.FileRecords;
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MemoryRecordsBuilder;
+import org.apache.kafka.common.record.TimestampType;
+
+import org.assertj.core.util.Files;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class SegmentCompressionCheckerTest {
+
+    @TempDir
+    Path dir;
+
+    @Test
+    void shouldFailWhenReadingEmptyFile() {
+        assertThatThrownBy(() -> SegmentCompressionChecker.check(Files.newTemporaryFile()))
+            .isInstanceOf(InvalidRecordBatchException.class)
+            .hasMessage("Record batch is null");
+    }
+
+    @Test
+    void shouldFailWhenReadingInvalidFile() throws IOException {
+        // Given a valid record batch
+        final Path path = dir.resolve("invalid.log");
+        final File file = Files.newFile(path.toString());
+        try (final FileRecords records = FileRecords.open(file, false, 100000, true);
+             final MemoryRecordsBuilder builder = MemoryRecords.builder(
+                 ByteBuffer.allocate(1024),
+                 CompressionType.NONE,
+                 TimestampType.CREATE_TIME,
+                 0)) {
+            builder.append(0L, "key-0".getBytes(), "value-0".getBytes());
+            records.append(builder.build());
+        }
+        // When messing with the record content
+        final byte[] bytes = java.nio.file.Files.readAllBytes(path);
+        final byte[] output = new byte[bytes.length];
+        for (int i = 0; i < bytes.length; i++) {
+            if (bytes[i] == "0".getBytes()[0]) {
+                output[i] = "1".getBytes()[0];
+            } else {
+                output[i] = bytes[i];
+            }
+        }
+        java.nio.file.Files.write(path, output);
+        // Then
+        assertThatThrownBy(() -> SegmentCompressionChecker.check(file))
+            .isInstanceOf(InvalidRecordBatchException.class)
+            .hasMessage("Failed to read and validate first batch");
+    }
+
+    @ParameterizedTest
+    @CsvSource({"NONE,false", "ZSTD,true"})
+    void shouldReturnCompressedWhenEnabled(final CompressionType compressionType, final boolean result)
+        throws InvalidRecordBatchException, IOException {
+        final File file = Files.newFile(dir.resolve("segment.log").toString());
+        try (final FileRecords records = FileRecords.open(file, false, 100000, true);
+             final MemoryRecordsBuilder builder = MemoryRecords.builder(
+                 ByteBuffer.allocate(1024),
+                 compressionType,
+                 TimestampType.CREATE_TIME,
+                 0)) {
+            builder.append(0L, "key-0".getBytes(), "value-0".getBytes());
+            records.append(builder.build());
+        }
+
+        final boolean requires = SegmentCompressionChecker.check(file);
+        assertThat(requires).isEqualTo(result);
+    }
+
+    @ParameterizedTest
+    @CsvSource({"NONE,ZSTD,false", "ZSTD,NONE,true"})
+    void shouldReturnCompressedWhenCompressionChanges(final CompressionType firstCompressionType,
+                                                      final CompressionType nextCompressionType,
+                                                      final boolean result)
+        throws InvalidRecordBatchException, IOException {
+        final File file = Files.newFile(dir.resolve("segment.log").toString());
+        try (final FileRecords records = FileRecords.open(file, false, 100000, true);
+             final MemoryRecordsBuilder b1 = MemoryRecords.builder(
+                 ByteBuffer.allocate(1024),
+                 firstCompressionType,
+                 TimestampType.CREATE_TIME,
+                 0);
+             final MemoryRecordsBuilder b2 = MemoryRecords.builder(
+                 ByteBuffer.allocate(1024),
+                 nextCompressionType,
+                 TimestampType.CREATE_TIME,
+                 0)) {
+            b1.append(0L, "key-0".getBytes(), "value-0".getBytes());
+            records.append(b1.build());
+            b2.append(0L, "key-0".getBytes(), "value-0".getBytes());
+            records.append(b2.build());
+        }
+
+        final boolean requires = SegmentCompressionChecker.check(file);
+        assertThat(requires).isEqualTo(result);
+    }
+}
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/InvalidRecordBatchException.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/InvalidRecordBatchException.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage.commons;
+
+public class InvalidRecordBatchException extends Exception {
+    public InvalidRecordBatchException(final String message) {
+        super(message);
+    }
+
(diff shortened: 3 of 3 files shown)


commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,55 @@
...
+                builder.append(0L, ("key-0").getBytes(), ("value-0").getBytes());
+                records.append(builder.build());
+            }
+        }
+
+        final boolean requires = new SegmentCompressionChecker().apply(file);
+        assertFalse(requires);
+    }
+
+}
```
ivanyu, line 55: Nit: new line

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MemoryRecordsBuilder;
+import org.apache.kafka.common.record.TimestampType;
+import org.assertj.core.util.Files;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class SegmentCompressionCheckerTest {
```
ivanyu, line 18: It would be great to have a test for cases where the first and second batches have different compression.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,28 @@
+package io.aiven.kafka.tieredstorage.commons.io;
```
ivanyu, line 1: Not sure it's `io`. I'd put it directly to `commons`.

ivanyu, line 1: The copyright header is missing (Checkstyle complains about this, too).

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,28 @@
...
+ * To be used when segment files are received on archival.
+ */
+public class SegmentCompressionChecker implements Function<File, Boolean> {
+
+    /**
+     * @param file Kafka log segment file
+     * @return true if requires compression, else false
+     */
+    @Override
+    public Boolean apply(File file) {
```
ivanyu, line 20:
```suggestion
    public boolean apply(final File file) {
```

Also, maybe some more concrete name, like `check`?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,28 @@
...
+        try (final FileRecords records = FileRecords.open(file)){
+            return records.firstBatch().compressionType() == CompressionType.NONE;
+        } catch (IOException e) {
```
ivanyu, line 23: Just to check: what does Kafka throw on various validation errors, e.g. CRC mismatch?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,55 @@
...
+                TimestampType.CREATE_TIME,
+                0
+            )) {
+                builder.append(0L, ("key-0").getBytes(), ("value-0").getBytes());
+                records.append(builder.build());
+            }
+        }
+
+        final boolean requires = new SegmentCompressionChecker().apply(file);
+        assertFalse(requires);
```
ivanyu, line 52: Please switch to AssertJ checks.

ivanyu (changes requested):
Some comments

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
jeqo: With current approach (take only first batch) this wouldn't exercise any additional paths, though (ie. checker will return false) Unless you want to change the approach to compare first and last batch?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
jeqo: it's like this because it is implementing `Function<File, Boolean>`. Should we fallback to plain class?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
jeqo: Good question. For safety we may extend the catch to Exception, and make a note on it, wdyt?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.kafka.common.record.FileRecords;
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MemoryRecordsBuilder;
+import org.apache.kafka.common.record.TimestampType;
+
+import org.assertj.core.util.Files;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
```
ivanyu, line 34: I guess it should be
```suggestion
import static org.assertj.core.api.Assertions.assertThat;
```
?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright 2021 Aiven Oy
```
ivanyu, line 2:
```suggestion
 * Copyright 2023 Aiven Oy
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2021 Aiven Oy
```
ivanyu, line 2:
```suggestion
 * Copyright 2023 Aiven Oy
```

ivanyu (changes requested):

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
ivanyu: Yeah, makes sense. Probably also some logging wouldn't be bad.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
ivanyu: Yeah, probably `Function` doesn't give us much

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
ivanyu: I'm more about making the contract more explicit: to show only the first batch matters.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionChecker.java
jeqo: Sure, adding logging.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,45 @@
...
+     * @param file Kafka log segment file
+     * @return true if requires compression, else false
+     */
+    @Override
+    public Boolean apply(final File file) {
+        try (final FileRecords records = FileRecords.open(file)) {
+            return records.firstBatch().compressionType() == CompressionType.NONE;
+        } catch (final IOException e) {
+            // Swallow exception as it should not interrupt processing
+            return false;
```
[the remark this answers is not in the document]
jeqo: Agree. Not much we can do at this stage to validate the segment file. I'm ok with either returning false or throwing a RuntimeException, wdyt?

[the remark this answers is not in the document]
ivanyu: Or maybe a special exception type for this?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/io/SegmentCompressionCheckerTest.java
jeqo: I see! Make sense, adding these.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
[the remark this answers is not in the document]
jeqo: Not sure. We may as well just throw IOException and leave it to the client to handle it?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class SegmentCompressionChecker {
+
+    /**
+     * @param file Kafka log segment file
+     * @return true if requires compression, else false
+     */
+    public static boolean check(final File file) throws IOException {
+        try (final FileRecords records = FileRecords.open(file)) {
+            return records.firstBatch().compressionType() == CompressionType.NONE;
```
ivanyu, line 37: I think we also need to call `ensureValid()` on the batch.

It throws `CorruptRecordException` in case the batch is invalid. We need to think how to propagate it further (it's unchecked).

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class SegmentCompressionCheckerTest {
+
+    @TempDir
+    Path dir;
+
+    @Test
+    void shouldRequireCompression_whenCompressionNone() throws IOException {
```
AnatolyPopov, line 42: Should we have 2 parameterized tests one for case when compression changes and one when it does not?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
jeqo: Not sure about calling `ensureValid` at this point. It will validate CRC introducing additional overhead, while Compression type is on the record batch headers already. 
If we are looking to ensure a log segment is valid, we could do it outside this compression check and pass FileRecords instead of File as input, wdyt?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,133 @@
...
+        final File file = Files.newFile(dir.resolve("segment.log").toString());
+        try (final FileRecords records = FileRecords.open(file, false, 100000, true)) {
+            try (final MemoryRecordsBuilder builder = MemoryRecords.builder(
```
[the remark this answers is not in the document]
jeqo: With try-with-resource I don't see the issue having nested `try`s.
Are you suggesting to put all resources on the same try? It may save us a line or two but not sure is more readable.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
jeqo: Not sure the complexity of the tests justify parametrized tests at this point. IIUC, It would require passing compression type + expected output as parameters which adds overhead when reading the tests. 
Though, if I'm missing some scenarios we could consider it.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
ivanyu: > If we are looking to ensure a log segment is valid

No, that's not the point. The idea is that since we're reading from disk, any corruption is possible and if the batch is invalid, `compressionType` may contain nonsense (may be just incorrect or throw `IllegalArgumentException` because the value doesn't match anything known). `ensureValid` is the way to ensure the batch integrity with good certainty.
I wouldn't bother of the overhead also, CRC32 calculation is fast and this segment will reach the page cache anyway.

jeqo: Sure, fair enough. Giving a try now, lmk how it looks

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+        java.nio.file.Files.write(path, output);
+        // Then
+        try {
+            SegmentCompressionChecker.check(file);
+            fail("Expected to fail");
+        } catch (InvalidRecordBatchException e) {
+            assertThat(e).hasMessage("Failed to read and validate first batch");
+        }
```
ivanyu, line 89: I think it should be something like
```suggestion
        assertThatThrownBy(() -> SegmentCompressionChecker.check(file))
            .isInstanceOf(InvalidRecordBatchException.class)
            .hasMessage("Failed to read and validate first batch");
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Checks if segment requires compression or not.
+ * To be used when segment files are received on archival.
+ */
+public class SegmentCompressionChecker {
+
+    /**
+     * @param file Kafka log segment file
+     * @return true if requires compression, else false
```
ivanyu, line 35: Isn't this a bit counter-intuitive? One would expect that a compression checker returns `true` when it detects the segment **is** compressed, not otherwise. Maybe flip?

ivanyu (changes requested):
A couple of comments + Checkstlyle complains

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
jeqo: Sure!

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
jeqo: On a second thought, I agree parametrized tests could help here. I gave a try, lmk how it looks.

AnatolyPopov: LGTM

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionCheckerTest.java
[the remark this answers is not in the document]
jeqo: I gave a try, lmk if looks good.

ivanyu (approved):
LGTM

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @param file Kafka log segment file
+     * @return true if log segment is compressed, otherwise returns false
+     */
+    public static boolean check(final File file) throws InvalidRecordBatchException {
+        try (final FileRecords records = FileRecords.open(file, false, true, 0, false)) {
+            final RecordBatch batch = fistRecordBatch(records);
+            return batch.compressionType() != CompressionType.NONE;
+        } catch (final IOException | CorruptRecordException e) {
```
AnatolyPopov, line 41: Should we also catch `KafkaException` here? If file that we are trying to read is a directory for some reason(at least) `RecordBatchIterator`throws it. In that case we will fail anyways and the file will never be uploaded but I wonder if we should wrap into our exception or propagate further?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import org.apache.kafka.common.errors.CorruptRecordException;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.FileRecords;
+import org.apache.kafka.common.record.RecordBatch;
+
+/**
+ * Checks if segment requires compression or not.
```
AnatolyPopov, line 28: But are not we checking here if it's already compressed and not making a decision here if it needs to be compressed? I think we should rephrase this a little.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
jeqo: Good catch! Sure, we can replace CorruptRecordException with KafkaException (super class) and wrap them all

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentCompressionChecker.java
jeqo: Agree, fixing it now

AnatolyPopov:
@jeqo  can you squash?

AnatolyPopov (approved):
LGTM

PR #175, opened 2023-04-20
UniversalRemoteStorageManager implementation

AnatolyPopov:
Adding RemoteStorageManager implementation that supposed to be used as a single entry point all possible future remote storage implementations.

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
@@ -16,54 +16,223 @@
 
 package io.aiven.kafka.tieredstorage.commons;
 
+import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
+
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.nio.file.Files;
 import java.util.Map;
+import java.util.Objects;
 
+import org.apache.kafka.common.utils.ByteBufferInputStream;
 import org.apache.kafka.server.log.remote.storage.LogSegmentData;
 import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
 import org.apache.kafka.server.log.remote.storage.RemoteStorageException;
 import org.apache.kafka.server.log.remote.storage.RemoteStorageManager;
 
+import io.aiven.kafka.tieredstorage.commons.manifest.SegmentEncryptionMetadataV1;
+import io.aiven.kafka.tieredstorage.commons.manifest.SegmentManifest;
+import io.aiven.kafka.tieredstorage.commons.manifest.SegmentManifestV1;
+import io.aiven.kafka.tieredstorage.commons.manifest.index.ChunkIndex;
+import io.aiven.kafka.tieredstorage.commons.manifest.serde.DataKeyDeserializer;
+import io.aiven.kafka.tieredstorage.commons.manifest.serde.DataKeySerializer;
+import io.aiven.kafka.tieredstorage.commons.security.AesEncryptionProvider;
+import io.aiven.kafka.tieredstorage.commons.security.DataKeyAndAAD;
+import io.aiven.kafka.tieredstorage.commons.security.RsaEncryptionProvider;
+import io.aiven.kafka.tieredstorage.commons.storage.ObjectStorageFactory;
+import io.aiven.kafka.tieredstorage.commons.transform.BaseTransformChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.CompressionChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.EncryptionChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.FetchChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.TransformChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.TransformFinisher;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.LEADER_EPOCH;
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.OFFSET;
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.PRODUCER_SNAPSHOT;
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.TIMESTAMP;
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.TRANSACTION;
+
 public class UniversalRemoteStorageManager implements RemoteStorageManager {
+    private UniversalRemoteStorageManagerConfig config;
+
+    private ObjectStorageFactory objectStorageFactory;
+    private boolean compression;
+    private boolean encryption;
+    private int chunkSize;
+    private RsaEncryptionProvider rsaEncryptionProvider;
+    private AesEncryptionProvider aesEncryptionProvider;
+    private ObjectMapper mapper;
+    private ChunkManager chunkManager;
+    private ObjectKey objectKey;
+
     @Override
     public void configure(final Map<String, ?> configs) {
-        throw new UnsupportedOperationException();
+        Objects.requireNonNull(configs, "configs must not be null");
+        config = new UniversalRemoteStorageManagerConfig(configs);
+        objectStorageFactory = config.objectStorageFactory();
+        objectKey = new ObjectKey(config.keyPrefix());
+        encryption = config.encryptionEnabled();
+        if (encryption) {
+            rsaEncryptionProvider = RsaEncryptionProvider.of(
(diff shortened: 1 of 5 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/manifest/index/AbstractChunkIndex.java
```diff
@@ -91,7 +91,7 @@ public Chunk findChunkForOriginalOffset(final int offset) {
             //                ^
             //             offset
             final int firstOffsetBeyondCurOriginalChunk = (chunkI + 1) * originalChunkSize;
-            if (offset < firstOffsetBeyondCurOriginalChunk) {
+            if (offset <= firstOffsetBeyondCurOriginalChunk) {
```
ivanyu, line 94: To warn in advance: this seems an incorrect change. It breaks the `VariableSizeChunkIndexBuilderTest` and `VariableSizeChunkIndexBuilderTest` tests and it's a legit failure.

AnatolyPopov: Yep, fixed

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
```diff
@@ -16,54 +16,263 @@
...
     public InputStream fetchIndex(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                                   final IndexType indexType) throws RemoteStorageException {
-        throw new UnsupportedOperationException();
+        try {
+            return objectStorageFactory.fileFetcher()
+                .fetch(objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.fromIndexType(indexType)));
+        } catch (final IOException e) {
+            throw new RemoteStorageException(e);
+        }
+
```
ivanyu, line 259: I tried running this against the current TS broker implementation and it fails as it tries to fetch `IndexType.TRANSACTION`, which may or may not exist. It seems, it expects `null` if it doesn't exist:
https://github.com/aiven/kafka/blob/70c4fb03da91ba8c9b74af2200696e7db5347c6c/core/src/main/scala/kafka/log/remote/RemoteIndexCache.scala#L185

I think we should return `null` (if there's indeed no transaction index) and also leave a TODO comment here (or a ticket) that this needs to be aligned with the future upstream implementation.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
```diff
@@ -16,54 +16,263 @@
...
+            final Cipher encryptCipher = getCipher();
+            encryptCipher.init(Cipher.DECRYPT_MODE, encryptionMetadata.secretKey(),
+                new IvParameterSpec(encryptedChunk, 0, encryptionMetadata.ivSize()),
+                SecureRandom.getInstanceStrong());
+            encryptCipher.updateAAD(encryptionMetadata.aad());
+            return encryptCipher;
+        } catch (final NoSuchAlgorithmException | InvalidKeyException | InvalidAlgorithmParameterException e) {
+            throw new RuntimeException(e);
+        }
     }
```
jeqo, line 223: This doesn't belong here. Should we move it to `security.EncryptionKeyProvider`?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
ivanyu: The doc is not that clear:
```RemoteResourceNotFoundException – when there are no resources associated with the given remoteLogSegmentMetadata```
I read it that way that "no resources" = "the segment doesn't exist at all". This may matter because in the upstream implementation (now or at some point) some decisions that we don't want may be made based on this exception

File: commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java (hunk 3 above)
Comment (jeqo) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java]: Solved on #184

**Comment by jeqo** [comment]
Rebased to latest changes, RSM tests still green

File: commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java (hunk 2 above)
Comment (AnatolyPopov) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: We hsould decide something on this. Looks like @jeqo is right here but on the other hand it is unclear how to distinguish lack of transaction index when it is not configured vs missing transaction issues in case of any issues.

Comment (jeqo) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: Looking for clarification here: https://issues.apache.org/jira/browse/KAFKA-7739?focusedCommentId=17720936&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17720936

Comment (ivanyu) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: Return `null` + leave a TODO for now?

Comment (jeqo) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: Interestingly, upstream doesn't handle this exception as in the linkedin fork: https://github.com/apache/kafka/blob/322ac86ba282f35373382854cc9e790e4b7fb5fc/core/src/main/scala/kafka/log/remote/RemoteIndexCache.scala#L197-L211
and if returning null, it will break as exception is not handled AFAICS.

Comment (AnatolyPopov) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: In the LInkedIn fork AFAIU `remoteStorageManager.fetchIndex` is wrapped into `Option` for that specific case(e.g. null was returned)

Comment (AnatolyPopov) [reply commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java:214]: Seems if we are sticking to LinkedIn version for now then null should probably be fine. So proceeding with @ivanyu's suggestion

File: commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java [hunk 4]
```diff
@@ -16,54 +16,223 @@
...
+                final String fileKey =
+                    objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.LOG);
+                objectStorageFactory.fileUploader().upload(sis, fileKey);
+            }
+
+
+            final ChunkIndex chunkIndex = transformFinisher.chunkIndex();
+            final SegmentManifest segmentManifest =
+                new SegmentManifestV1(chunkIndex, compression, encryptionMetadata);
+            uploadManifest(remoteLogSegmentMetadata, segmentManifest);
```
jeqo, line 141: minor: could we move this into its own method? maybe `uploadLogSegmentAndManifest` or similar?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerTest.java
```diff
@@ -77,6 +83,8 @@ class UniversalRemoteStorageManagerTest extends RsaKeyAwareTest {
     Path producerSnapshotFilePath;
     Path txnIndexFilePath;
 
+    public static final byte[] LEADER_EPOCH_INDEX_BYTES = "leader epoch index".getBytes();
+    static final int IV_SIZE = 12;
```
jeqo, line 87: `SegmentEncryptionMetadataV1` already has this constant. Maybe reuse it? We can move it later to the encryption provider where it should belong I think.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerTest.java
```diff
@@ -323,16 +338,24 @@ private void checkFetching(final int chunkSize) throws RemoteStorageException, I
...
-
-                    assertThat(rsm.fetchLogSegment(REMOTE_LOG_METADATA, offset, offset + readSize))
-                        .hasSameContentAs(expectedInputStream);
+        for (final int readSize : List.of(1, 13, actualChunkSize / 2, actualChunkSize, actualChunkSize + 1,
+            actualChunkSize * 2)) {
+            final byte[] expectedBytes = new byte[readSize];
+            for (final int offset : List.of(0, 1, 23, actualChunkSize / 2, actualChunkSize - 1, actualChunkSize,
+                actualChunkSize + 1
+//                ,
+//                SEGMENT_SIZE - readSize - 10
```
jeqo, line 349: rm this?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
AnatolyPopov: I was thinking that we will make the refactoring that we were planning with you(the pipelines) and refactor this at that time. Moving it to a separate method before does not bring too much value IMO but keeping it here can work as a reminder :)

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerTest.java
AnatolyPopov: Done

jeqo (approved):

ivanyu (approved):

ivanyu:
🎉

PR #178, opened 2023-04-20
fix: remove parent directories to mimic s3 behaviour

jeqo:
When a directory is emptied on S3, the parent keys are removed until an object is found

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
@@ -74,8 +74,8 @@ class FileSystemStorage implements FileUploader, FileFetcher, FileDeleter {
 
         final Path path = fsRoot.resolve(key);
         final long fileSize = Files.size(path);
-        if (to > fileSize) {
-            throw new IllegalArgumentException("to cannot be bigger than file size, to="
+        if (to >= fileSize) {
+            throw new IllegalArgumentException("position 'to' cannot be equal or higher than the file size, to="
                 + to + ", file size=" + fileSize + " given");
         }
 
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
@@ -26,20 +26,28 @@ import io.aiven.kafka.tieredstorage.commons.storage.FileDeleter;
 import io.aiven.kafka.tieredstorage.commons.storage.FileFetcher;
 import io.aiven.kafka.tieredstorage.commons.storage.FileUploader;
 
+import org.apache.commons.io.file.PathUtils;
 import org.apache.commons.io.input.BoundedInputStream;
 
 class FileSystemStorage implements FileUploader, FileFetcher, FileDeleter {
-    private final String fsRoot;
+    private final Path fsRoot;
     private final boolean overwrites;
 
     FileSystemStorage(final String fsRoot, final boolean overwrites) {
+        this(Path.of(fsRoot), overwrites);
+    }
+
+    FileSystemStorage(final Path fsRoot, final boolean overwrites) {
+        if (!Files.isDirectory(fsRoot) || !Files.isWritable(fsRoot)) {
+            throw new IllegalArgumentException(fsRoot + " must be a writable directory");
+        }
         this.fsRoot = fsRoot;
         this.overwrites = overwrites;
     }
 
     @Override
     public void upload(final InputStream inputStream, final String key) throws IOException {
-        final Path path = Path.of(fsRoot, key);
+        final Path path = fsRoot.resolve(key);
         if (!overwrites && Files.exists(path)) {
             throw new IOException("File " + path + " already exists");
         }
@@ -51,7 +59,7 @@ class FileSystemStorage implements FileUploader, FileFetcher, FileDeleter {
 
     @Override
     public InputStream fetch(final String key) throws IOException {
-        final Path path = Path.of(fsRoot, key);
+        final Path path = fsRoot.resolve(key);
         return Files.newInputStream(path);
     }
 
@@ -64,19 +72,28 @@ class FileSystemStorage implements FileUploader, FileFetcher, FileDeleter {
             throw new IllegalArgumentException("from cannot be more than to, from=" + from + ", to=" + to + " given");
         }
 
-        final Path path = Path.of(fsRoot, key);
+        final Path path = fsRoot.resolve(key);
         final long fileSize = Files.size(path);
         if (to > fileSize) {
             throw new IllegalArgumentException("to cannot be bigger than file size, to="
                 + to + ", file size=" + fileSize + " given");
         }
-        final InputStream result = new BoundedInputStream(Files.newInputStream(path), to + 1);
+
+        final int size = to + 1;
+        final InputStream result = new BoundedInputStream(Files.newInputStream(path), size);
         result.skip(from);
         return result;
     }
 
     @Override
     public void delete(final String key) throws IOException {
-        Files.delete(Path.of(fsRoot, key));
+        final Path path = fsRoot.resolve(key);
+        Files.deleteIfExists(path);
+        Path parent = path.getParent();
+        while (parent != null && Files.isDirectory(parent) && !parent.equals(fsRoot)
+            && PathUtils.isEmptyDirectory(parent)) {
+            Files.deleteIfExists(parent);
+            parent = parent.getParent();
+        }
     }
 }
--- a/commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
+++ b/commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage.commons.storage.filesystem;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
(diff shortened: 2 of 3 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -77,6 +78,16 @@ public InputStream fetch(final String key, final int from, final int to) throws
...
     public void delete(final String key) throws IOException {
-        Files.delete(Path.of(fsRoot, key));
+        final Path path = Path.of(fsRoot, key);
+        Files.deleteIfExists(path);
+        Path parent = path.getParent();
+        while (parent != null && Files.isDirectory(parent) && !parent.equals(Path.of(fsRoot))) {
+            try (final DirectoryStream<Path> directory = Files.newDirectoryStream(parent)) {
+                if (!directory.iterator().hasNext()) {
+                    Files.deleteIfExists(parent);
+                }
```
ivanyu, line 88: I think here we can `break` in `else`.

ivanyu (changes requested):
A small comment + some tests would be nice

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
jeqo: Good catch!

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+        Files.createDirectories(parentPath);
+        Files.writeString(parentPath.resolve("another"), "test");
+        final Path keyPath = parentPath.resolve(key);
+        Files.writeString(keyPath, "test");
+        final FileSystemStorage storage = new FileSystemStorage(root, false);
+        storage.delete(parent + "/" + key);
+
+        assertFalse(Files.exists(keyPath));
+        assertTrue(Files.exists(parentPath));
+        assertTrue(Files.exists(root));
```
AnatolyPopov, line 179: Lets stick to assrtj where it is possible.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -64,19 +72,33 @@ public InputStream fetch(final String key, final int from, final int to) throws
...
+        while (parent != null && Files.isDirectory(parent) && !parent.equals(fsRoot)) {
+            try (final DirectoryStream<Path> directory = Files.newDirectoryStream(parent)) {
+                if (!directory.iterator().hasNext()) {
+                    Files.deleteIfExists(parent);
+                } else {
+                    break;
+                }
+            }
+            parent = parent.getParent();
+        }
```
AnatolyPopov, line 102: I just realized that we have commons-io in the classpath already so maybe this? 
```suggestion
       while (parent != null && Files.isDirectory(parent) && !parent.equals(fsRoot)
            && PathUtils.isEmptyDirectory(parent)) {
            Files.deleteIfExists(parent);
            parent = parent.getParent();
        }
```

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class FileSystemStorageTest {
+
+    @Test
+    void testRootCannotBeAFile(@TempDir final Path root) throws IOException {
```
AnatolyPopov, line 41: Can we move this repeating argument into a field?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+        final String content = "content";
+        final String key = "key";
+        storage.upload(new ByteArrayInputStream(content.getBytes()), key);
+
+        assertEquals(content, Files.readString(root.resolve(key)));
+    }
+
+    @Test
+    void testUploadFailsWhenFileExists(@TempDir final Path root) throws IOException {
+        final String key = "key";
```
AnatolyPopov, line 74: Would it make sense to create constant for this dummy key instead of creating a variable in almost every test method?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+
+        try (final InputStream fetch = storage.fetch(key)) {
+            assertArrayEquals(content.getBytes(), fetch.readAllBytes());
+        }
+    }
+
+    @Test
+    void testFetchWithOffsetRange(@TempDir final Path root) throws IOException {
+        final String key = "key";
+        final String content = "content";
```
AnatolyPopov, line 107: Same as for key, maybe constant?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class FileSystemStorageTest {
```
AnatolyPopov, line 38: I believe one of the main cases is missing: key containing `/`, could you please add it or eve change a default key to something containing `/`?

AnatolyPopov (changes requested):
Some test-related comments

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
jeqo: Sure! Good catch

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+            assertArrayEquals(prefix.getBytes(), fetch.readAllBytes());
+        }
+    }
+
+    @Test
+    void testFetchWithFailsWithWrongOffsets(@TempDir final Path root) {
+        final FileSystemStorage storage = new FileSystemStorage(root, true);
+
+        assertThrows(IllegalArgumentException.class, () -> storage.fetch("key", -1, 1));
+        assertThrows(IllegalArgumentException.class, () -> storage.fetch("key", 2, 1));
```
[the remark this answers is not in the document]
jeqo: Sure, updated version with assertj should cover this.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
jeqo: Good idea. Adding this to the new constant for sample keys.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
jeqo: Sure, creating a constant for this representing the topic/partition/segment structure to be used.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
jeqo: I guess this could be specific for the test case. Don't want to couple them to the content value.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -0,0 +1,181 @@
...
+        try (final InputStream fetch = storage.fetch(key)) {
+            assertArrayEquals(content.getBytes(), fetch.readAllBytes());
+        }
+    }
+
+    @Test
+    void testFetchWithOffsetRange(@TempDir final Path root) throws IOException {
+        final String key = "key";
+        final String content = "content";
+        final String prefix = content.substring(3);
```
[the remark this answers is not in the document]
jeqo: Initially it was 0,3 (prefix) but forgot renaming 😅 , thanks for spotting this!

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -64,19 +72,28 @@ public InputStream fetch(final String key, final int from, final int to) throws
             throw new IllegalArgumentException("from cannot be more than to, from=" + from + ", to=" + to + " given");
         }
 
-        final Path path = Path.of(fsRoot, key);
+        final Path path = fsRoot.resolve(key);
         final long fileSize = Files.size(path);
         if (to > fileSize) {
```
AnatolyPopov, line 77: Seems we have a bug here, since we should still throw an exception if `to == fileSize` and we need a test for this. Since you are adding all kinds of tests in this PR can you also fix this and add a test?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -64,19 +72,33 @@ public InputStream fetch(final String key, final int from, final int to) throws
...
             throw new IllegalArgumentException("to cannot be bigger than file size, to="
                 + to + ", file size=" + fileSize + " given");
         }
```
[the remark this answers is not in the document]
AnatolyPopov: Good question but I believe we should not rely on this and this should be an error, it's better to fail fast then in the end trying to figure out why a part of the batch is missing.

AnatolyPopov (approved):
LGTM

PR #187, opened 2023-04-21
fix: chunk manager: return input stream instead of future

jeqo:
As Tiered Storage framework already covers the asynchronous calls to RSM, the Chunk Manager can return content as Input Stream directly.


AnatolyPopov:
Lets postpone this one a little bit because when we will implement the caching it might change once again and it's not absolutely clear in what way yet.

AnatolyPopov:
Actually looking into some other parts I've reconsidered it and I think we can merge this one. Can you @jeqo rebase to fix the tests?

AnatolyPopov (approved):
LGTM

PR #190, opened 2023-04-23
refactor: aes and rsa keys initialization

jeqo:
Changes:
- Moves RSA Path into InputStream mapping into one place
- Creates an AES Encryption Provider out of an RSA Encryption provider.

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/AesEncryptionProvider.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/AesEncryptionProvider.java
@@ -22,19 +22,32 @@ import javax.crypto.SecretKey;
 import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 
-import io.aiven.kafka.tieredstorage.commons.manifest.SegmentEncryptionMetadata;
+import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.SecureRandom;
 
-import static io.aiven.kafka.tieredstorage.commons.security.RsaEncryptionProvider.KEY_SIZE;
+import io.aiven.kafka.tieredstorage.commons.manifest.SegmentEncryptionMetadata;
 
 public class AesEncryptionProvider implements Encryption, Decryption {
 
+    public static final int KEY_SIZE = 512;
     public static final int KEY_AND_AAD_SIZE_BYTES = KEY_SIZE / 8 / 2;
     public static final String AES_TRANSFORMATION = "AES/GCM/NoPadding";
 
     private final KeyGenerator aesKeyGenerator;
 
-    public AesEncryptionProvider(final KeyGenerator aesKeyGenerator) {
-        this.aesKeyGenerator = aesKeyGenerator;
+    static KeyGenerator keyGenerator() {
+        try {
+            final KeyGenerator kg = KeyGenerator.getInstance("AES", "BC");
+            kg.init(KEY_SIZE, SecureRandom.getInstanceStrong());
+            return kg;
+        } catch (final NoSuchAlgorithmException | NoSuchProviderException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public AesEncryptionProvider() {
+        this.aesKeyGenerator = keyGenerator();
     }
 
     public SecretKey createDataKey() {
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/RsaEncryptionProvider.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/RsaEncryptionProvider.java
@@ -18,20 +18,19 @@ package io.aiven.kafka.tieredstorage.commons.security;
 
 import javax.crypto.BadPaddingException;
 import javax.crypto.IllegalBlockSizeException;
-import javax.crypto.KeyGenerator;
 import javax.crypto.SecretKey;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.nio.file.Files;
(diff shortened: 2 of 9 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/RsaEncryptionProvider.java
```diff
@@ -56,8 +58,8 @@ private RsaEncryptionProvider(final KeyPair rsaKeyPair) {
         this.rsaKeyPair = rsaKeyPair;
     }
 
-    public static RsaEncryptionProvider of(final InputStream rsaPublicKey,
-                                           final InputStream rsaPrivateKey) {
+    public static RsaEncryptionProvider of(final Path rsaPublicKey,
+                                           final Path rsaPrivateKey) {
```
ivanyu, line 62: Considering https://github.com/aiven/tiered-storage-for-apache-kafka/issues/198, maybe we should keep `InputStream`?

jeqo: Good point. What about keeping Paths here, and leave constructor with InputStream instead? Then we can add another static `of` for getting PEM content directly from config, wdyt?

ivanyu: Yeah, makes sense

jeqo: Wasn't actually the constructor, but I kept the InputStream method on RsaKeysReader added a new one for Paths. LMK how it looks now

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/RsaEncryptionProvider.java
```diff
@@ -91,6 +92,14 @@ public byte[] decryptDataKey(final byte[] bytes) {
 
     static class RsaKeysReader {
 
+        static KeyPair readRsaKeyPair(final Path publicKeyPath, final Path privateKeyPath) {
+            try {
+                return readRsaKeyPair(Files.newInputStream(publicKeyPath), Files.newInputStream(privateKeyPath));
```
ivanyu, line 97: These `InputStream`s must be closed, i.e. need to be in `try`

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/RsaEncryptionProvider.java
```diff
@@ -91,6 +92,14 @@ public byte[] decryptDataKey(final byte[] bytes) {
...
+            } catch (IOException e) {
```
ivanyu, line 98:
```suggestion
            } catch (final IOException e) {
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/AesEncryptionProvider.java
```diff
@@ -33,7 +36,15 @@ public class AesEncryptionProvider implements Encryption, Decryption {
 
     private final KeyGenerator aesKeyGenerator;
 
-    public AesEncryptionProvider(final KeyGenerator aesKeyGenerator) {
+    public static AesEncryptionProvider of(final RsaEncryptionProvider rsaEncryptionProvider) {
+        try {
+            return new AesEncryptionProvider(rsaEncryptionProvider.keyGenerator());
+        } catch (NoSuchAlgorithmException | NoSuchProviderException e) {
```
ivanyu, line 42:
```suggestion
        } catch (final NoSuchAlgorithmException | NoSuchProviderException e) {
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/AesEncryptionProvider.java
```diff
@@ -33,7 +36,15 @@ public class AesEncryptionProvider implements Encryption, Decryption {
 
     private final KeyGenerator aesKeyGenerator;
 
-    public AesEncryptionProvider(final KeyGenerator aesKeyGenerator) {
+    public static AesEncryptionProvider of(final RsaEncryptionProvider rsaEncryptionProvider) {
```
ivanyu, line 39: I'm not sure the whole key generator code should at all be in `RsaEncryptionProvider` as it's not used there. This means, `AesEncryptionProvider` and `RsaEncryptionProvider` could be completely independent. The `KEY_SIZE` constant should also be in `AesEncryptionProvider`. Maybe I'm missing something that justifies the current coupling between the two classes?

ivanyu (changes requested):

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/security/AesEncryptionProvider.java
jeqo: Good point.
Even though we may want to keep these 2 decoupled, most users of these APIs have to initialize them together. I'm touching on that in #191 
Let me know if you feel we should decouple these here and discuss the coupling issues on #191

ivanyu: Sure, let's postpone it until #191

jeqo: Oh, I just realized, RSA is not required for AES at all! Let me refactor it based on this suggestions and remove the coupling then

ivanyu: Yeah, they meet each other only when we encrypt AES keys with RSA.

ivanyu (approved):

PR #200, opened 2023-05-03
Use cp-kafka as base for Docker image

ivanyu:


jeqo:
Changes look good to me. Could we also adapt the docker compose files to the new image?

ivanyu:
Yes, those needs updating too.

However, I faced an obstacle: `KAFKA_RSM_CONFIG_S3_PUBLIC_KEY_PEM` env variable converts into `rsm.config.s3.public.key.pem`, while the "old" S3 plugin expects `<...>s3.public_key_pem` (which breaks the convention). I don't think it makes sense to look for a workaround: in the new implementation we'll follow the convention. So, let's pause this PR for a while.

PR #204, opened 2023-05-04
Don't allow originalChunkSize to be 0

ivanyu:
This doesn't make sense and also cause division by 0 in `FixedSizeChunkIndex.chunkCount`


AnatolyPopov (approved):
LGTM

PR #206, opened 2023-05-07
Adding ChunkManager implementation

AnatolyPopov:
Adding ChunkManager implementation that allows to fetch a specific chunk and return as decrypted and decompressed stream of bytes.

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
@@ -18,18 +18,52 @@ package io.aiven.kafka.tieredstorage.commons;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.List;
 
 import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
 
 import io.aiven.kafka.tieredstorage.commons.manifest.SegmentManifest;
+import io.aiven.kafka.tieredstorage.commons.security.AesEncryptionProvider;
+import io.aiven.kafka.tieredstorage.commons.storage.ObjectStorageFactory;
+import io.aiven.kafka.tieredstorage.commons.transform.BaseDetransformChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.DecompressionChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.DecryptionChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.DetransformChunkEnumeration;
+import io.aiven.kafka.tieredstorage.commons.transform.DetransformFinisher;
 
-public interface ChunkManager {
+public class ChunkManager {
+    private final ObjectStorageFactory objectStorageFactory;
+    private final ObjectKey objectKey;
+    private final AesEncryptionProvider aesEncryptionProvider;
+
+    public ChunkManager(final ObjectStorageFactory objectStorageFactory, final ObjectKey objectKey,
+            final AesEncryptionProvider aesEncryptionProvider) {
+        this.objectStorageFactory = objectStorageFactory;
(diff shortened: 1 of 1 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
```diff
@@ -18,18 +18,52 @@
...
-public interface ChunkManager {
     /**
      * Gets a chunk of a segment.
      *
      * @return an {@link InputStream} of the chunk, plain text (i.e. decrypted and decompressed).
      */
-    InputStream getChunk(RemoteLogSegmentMetadata remoteLogSegmentMetadata,
-                         SegmentManifest manifest,
-                         int chunkId) throws IOException;
+    public InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata, final SegmentManifest manifest,
```
jeqo, line 51: Could we add some tests for getting chunk, e.g. pipeline ordering (decryption and then decompression)

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
```diff
@@ -18,18 +18,52 @@
...
+        DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(segmentFile, List.of(chunk));
+        if (manifest.encryption().isPresent()) {
+            detransformEnum = new DecryptionChunkEnumeration(detransformEnum, manifest.encryption().get().ivSize(),
+                    encryptedChunk -> aesEncryptionProvider.decryptionCipher(encryptedChunk,
+                            manifest.encryption().get()));
+        }
+        if (manifest.compression()) {
+            detransformEnum = new DecompressionChunkEnumeration(detransformEnum);
+        }
+        final var detransformFinisher = new DetransformFinisher(detransformEnum);
```
jeqo, line 66: IIRC, we haven't used `var` on src (maybe on test?).
I'd suggest sticking to not using `var` unless we follow this on all the implementation:
```suggestion
        final DetransformFinisher detransformFinisher = new DetransformFinisher(detransformEnum);
```

jeqo (commented):
Thanks @AnatolyPopov.
I think we need some tests for this component as well.
 
Also, some general comments:
The `getChunk` implementation (decrypt, decompress) corresponds to the uploading logic (compress, encrypt). These should be on the same place.
I see two alternatives to do this:
- URSM implementing this interface
- Extending this class to handle all interaction between Kafka TS ops and ObjectStorageFactory:
  - In this case, I'd rename the class to `ChunkedSegmentManager` or similar, and extend it to `fetchChunk`, `uploadSegment` (including upload log, indexes, and manifest) and `deleteSegment`. 

I'm leaning towards the second option. Though, regardless, agreeing on either should help to scope this and the following PRs better. wdyt?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/ChunkManager.java
AnatolyPopov: We did in some places and I don't think `var` makes sense in 100% of cases but occasionally makes sense. So I agree for now we can stick to not using vars further unless we will change our mind.

AnatolyPopov:
@jeqo I totally agree regarding regarding your comment about having encryption/decryption and compression/decompression in the same place and was thinking about refactoring but I believe this can be addressed in other PRs and requires some discussions. For now I would proceed almost as is to finally get the first version working and then refactor this.

jeqo (approved):
Sure, let's review it after URSM implementation is merged. Creating https://github.com/aiven/tiered-storage-for-apache-kafka/issues/207 to follow up.

jeqo:
@AnatolyPopov could you squash commits? I can merge it right after

AnatolyPopov:
@jeqo Done

PR #208, opened 2023-05-08
feat: add KeyNotFoundException

jeqo:
Storage backends usually return exceptions when keys/files are not found. e.g. AmazonS3Exception 404 for S3, NoSuchFileException for FileSystem, etc. To wrap these into a common exception, this PR proposes a "KeyNotFoundException".

This exception should be catched and mapped to `RemoteResourceNotFoundException`

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
@@ -20,12 +20,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 
 import io.aiven.kafka.tieredstorage.commons.storage.BytesRange;
 import io.aiven.kafka.tieredstorage.commons.storage.FileDeleter;
 import io.aiven.kafka.tieredstorage.commons.storage.FileFetcher;
 import io.aiven.kafka.tieredstorage.commons.storage.FileUploader;
+import io.aiven.kafka.tieredstorage.commons.storage.KeyNotFoundException;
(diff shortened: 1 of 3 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/FileFetcher.java
```diff
@@ -24,13 +24,13 @@ public interface FileFetcher {
      * Fetch file.
      * @param key file key.
      */
-    InputStream fetch(String key) throws IOException;
+    InputStream fetch(String key) throws KeyNotFoundException, IOException;
```
ivanyu, line 27: Should be also in the deleter?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -73,6 +77,9 @@ public InputStream fetch(final String key, final int from, final int to) throws
         }
 
         final Path path = fsRoot.resolve(key);
+        if (!Files.exists(path)) {
+            throw new KeyNotFoundException(this, key);
+        }
```
ivanyu, line 82: I think that instead of this check, we'd better catch `NoSuchFileException` on `Files.newInputStream`: this eliminates this tiny but still possible window of the file disappearing after the check and before the opening attempt or some strange subtle disparity between the existence checks in `Files.exists` and `Files.newInputStream`.

ivanyu (changes requested):

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/FileFetcher.java
jeqo: Checking idempotent expectations on RSM: https://issues.apache.org/jira/browse/KAFKA-7739?focusedCommentId=17722747&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17722747

ivanyu (approved):

PR #209, opened 2023-05-08
feat: add object range for fetcher

jeqo:
Range validation should be in one place instead of each storage implementation.

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/BytesRange.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/BytesRange.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/BytesRange.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.aiven.kafka.tieredstorage.commons.storage;
+
+/**
+ * Byte range with from and to edges; where to cannot be less than from.
+ * inclusive/exclusive semantics are up to the consumer.
+ *
+ * @see FileFetcher#fetch(String, BytesRange)
+ * @see org.apache.kafka.server.log.remote.storage.RemoteStorageManager#fetchLogSegment(
```
AnatolyPopov, line 24: Do we need to have this reference here? In RSM interface there is no BytesRange so IMO it looks a little confusing. I would avoid having all this usage links altogether since the class + javadoc should be enough to understand how to use it IMO.

AnatolyPopov (approved):
LGTM

PR #213, opened 2023-05-09
refactor: remove overwrite flag for FileSystem storage

jeqo:
File system storage backend doesn't require overwrite flag as retries from RemoteStorageManager should always overwrite segments.


AnatolyPopov (approved):
LGTM

PR #216, opened 2023-05-10
refactor: align file-system storage with s3 

jeqo:
Couple of changes to align file-system storage with S3 and other back-ends:

- Do not throw exception when range end is larger than file size, but when range start is (introduce an "InvalidRangeException").
- Use range as inclusive start and end. Kafka TS fetch range is inclusive and S3 GET range is also inclusive.  
 
Depends on #215

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
@@ -52,6 +52,7 @@ import io.aiven.kafka.tieredstorage.commons.manifest.serde.DataKeySerializer;
 import io.aiven.kafka.tieredstorage.commons.security.AesEncryptionProvider;
 import io.aiven.kafka.tieredstorage.commons.security.DataKeyAndAAD;
 import io.aiven.kafka.tieredstorage.commons.security.RsaEncryptionProvider;
+import io.aiven.kafka.tieredstorage.commons.storage.BytesRange;
 import io.aiven.kafka.tieredstorage.commons.storage.ObjectStorageFactory;
 import io.aiven.kafka.tieredstorage.commons.storage.StorageBackEndException;
 import io.aiven.kafka.tieredstorage.commons.transform.BaseTransformChunkEnumeration;
@@ -232,8 +233,11 @@ public class UniversalRemoteStorageManager implements RemoteStorageManager {
     @Override
     public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                                        final int startPosition) throws RemoteStorageException {
-        return this.fetchLogSegment(remoteLogSegmentMetadata, startPosition,
-            remoteLogSegmentMetadata.segmentSizeInBytes());
+        return this.fetchLogSegment(
+            remoteLogSegmentMetadata,
+            startPosition,
+            remoteLogSegmentMetadata.segmentSizeInBytes() - 1
+        );
     }
 
     @Override
@@ -242,12 +246,17 @@ public class UniversalRemoteStorageManager implements RemoteStorageManager {
                                        final int endPosition) throws RemoteStorageException {
         try {
             final SegmentManifest segmentManifest = segmentManifestProvider.get(remoteLogSegmentMetadata);
+
+            final BytesRange range = BytesRange.of(
+                startPosition,
+                Math.min(endPosition, remoteLogSegmentMetadata.segmentSizeInBytes() - 1)
+            );
             final FetchChunkEnumeration fetchChunkEnumeration = new FetchChunkEnumeration(
                 chunkManager,
                 remoteLogSegmentMetadata,
                 segmentManifest,
-                startPosition,
-                endPosition);
+                range
+            );
             return new SequenceInputStream(fetchChunkEnumeration);
         } catch (final StorageBackEndException | IOException e) {
             throw new RemoteStorageException(e);
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
@@ -27,6 +27,7 @@ import io.aiven.kafka.tieredstorage.commons.storage.BytesRange;
 import io.aiven.kafka.tieredstorage.commons.storage.FileDeleter;
 import io.aiven.kafka.tieredstorage.commons.storage.FileFetcher;
 import io.aiven.kafka.tieredstorage.commons.storage.FileUploader;
+import io.aiven.kafka.tieredstorage.commons.storage.InvalidRangeException;
 import io.aiven.kafka.tieredstorage.commons.storage.KeyNotFoundException;
 import io.aiven.kafka.tieredstorage.commons.storage.StorageBackEndException;
 
@@ -73,14 +74,15 @@ class FileSystemStorage implements FileUploader, FileFetcher, FileDeleter {
         try {
             final Path path = fsRoot.resolve(key);
             final long fileSize = Files.size(path);
-            if (range.to > fileSize) {
-                throw new IllegalArgumentException("position 'to' cannot be higher than the file size, to="
-                    + range.to + ", file size=" + fileSize + " given");
+            if (range.from >= fileSize) {
+                throw new InvalidRangeException("Range start position " + range.from
+                    + " is outside file content. file size = " + fileSize);
             }
-
-            final InputStream result = new BoundedInputStream(Files.newInputStream(path), range.to);
-            result.skip(range.from);
-            return result;
+            // slice file content
+            final InputStream chunkContent = Files.newInputStream(path);
+            chunkContent.skip(range.from);
+            final long size = Math.min(range.to, fileSize) - range.from + 1;
+            return new BoundedInputStream(chunkContent, size);
         } catch (final NoSuchFileException e) {
             throw new KeyNotFoundException(this, key, e);
         } catch (final IOException e) {
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/transform/FetchChunkEnumeration.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/transform/FetchChunkEnumeration.java
@@ -29,6 +29,7 @@ import io.aiven.kafka.tieredstorage.commons.Chunk;
 import io.aiven.kafka.tieredstorage.commons.ChunkManager;
 import io.aiven.kafka.tieredstorage.commons.manifest.SegmentManifest;
 import io.aiven.kafka.tieredstorage.commons.manifest.index.ChunkIndex;
+import io.aiven.kafka.tieredstorage.commons.storage.BytesRange;
 import io.aiven.kafka.tieredstorage.commons.storage.StorageBackEndException;
 
 import org.apache.commons.io.input.BoundedInputStream;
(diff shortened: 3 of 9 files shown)


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -73,12 +74,21 @@ public InputStream fetch(final String key, final BytesRange range) throws Storag
         try {
             final Path path = fsRoot.resolve(key);
             final long fileSize = Files.size(path);
-            if (range.to > fileSize) {
-                throw new IllegalArgumentException("position 'to' cannot be higher than the file size, to="
-                    + range.to + ", file size=" + fileSize + " given");
+            if (fileSize > Integer.MAX_VALUE) {
+                throw new IOException("File size is larger than " + Integer.MAX_VALUE);
```
ivanyu, line 78: Do you think it should be an `IOException`, not `IllegalArgumentException` or `InvalidRangeException`?

ivanyu (changes requested):
> Do not throw exception when range end is larger than file size

Why may we want this? Doesn't this indicate that something is broken on the call site? I think we should be very explicit of this type of errors and fail when we detect them, otherwise we're creating a hell to debug

jeqo:
> Why may we want this? Doesn't this indicate that something is broken on the call site? 

Not really. At least from the API docs the end position may not be within file bounds:

```
     * Returns the remote log segment data file/object as InputStream for the given {@link RemoteLogSegmentMetadata}
     * starting from the given startPosition. The stream will end at the smaller of endPosition and the end of the
     * remote log segment data file/object.
```

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
jeqo: `fileSize` is not an input (to use IllegalArgExc), but a size larger than int bounds that may potentially break the casting, since nio.Files#size now return long instead of int. 
Also, not range to use the new InvalidRangeException.
So, IOException seems the closest exception, but there may be a better type?

AnatolyPopov: This IMO is in the wrong place. We are requesting the range and before we check that `to` is bigger than `fileSize` we should not care too much I guess since even if file is too big we can still read a part of it.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -116,22 +117,35 @@ void testFetchSingleByte() throws IOException, StorageBackEndException {
         Files.writeString(keyPath, content);
         final FileSystemStorage storage = new FileSystemStorage(root);
 
-        try (final InputStream fetch = storage.fetch(TOPIC_PARTITION_SEGMENT_KEY, BytesRange.of(2, 3))) {
+        try (final InputStream fetch = storage.fetch(TOPIC_PARTITION_SEGMENT_KEY, BytesRange.of(2, 2))) {
```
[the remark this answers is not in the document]
jeqo: start/end range positions are defined both as inclusive by RSM and at the storage layer. At least S3 refer to RFC 9110 Byte ranges [1]

> Examples of bytes range specifiers:
> - The first 500 bytes (byte offsets 0-499, inclusive):
>     bytes=0-499
> - The second 500 bytes (byte offsets 500-999, inclusive):
>     bytes=500-999

It also mentions:

> An [int-range](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range) is invalid if the [last-pos](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range) value is present and less than the [first-pos](https://www.rfc-editor.org/rfc/rfc9110.html#rule.int-range).

So, [2,2] should be valid and should return 1 byte.

GCS[2] and Azure[3] also mention both start/end should be included in response (ie. inclusive) or refer to HTTP range (RFC mentioned before)

[1] https://www.rfc-editor.org/rfc/rfc9110.html#name-byte-ranges
[2] https://cloud.google.com/storage/docs/samples/storage-download-byte-range#storage_download_byte_range-java
[3] https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-the-range-header-for-blob-service-operations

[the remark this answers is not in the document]
AnatolyPopov: Ah, sorry, seems because of GH folding I thought the comment from line 101 belongs to this test. You are correct.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
jeqo: @AnatolyPopov not really, as we are using the filesize to set the bounded inputstream later. The ranges expected by the framework as not considering files larger than max int--I guess as https://kafka.apache.org/documentation/#brokerconfigs_log.segment.bytes is defined as int-- so it may be worth to do this check here.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorageTest.java
```diff
@@ -96,8 +97,8 @@ void testFetchAll() throws IOException, StorageBackEndException {
     void testFetchWithOffsetRange() throws IOException, StorageBackEndException {
         final String content = "AABBBBAA";
         final int from = 2;
-        final int to = 6;
-        final String range = content.substring(from, to);
+        final int to = 5;
+        final String range = content.substring(from, to + 1); // exclusive -> inclusive
```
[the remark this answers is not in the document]
jeqo: Yeah, substring is exclusive, and turn in it into inclusive to match expected result. But you are right, let me make it more descriptive.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
AnatolyPopov: We are not using fileSize to set bounds for BoundedImputStream directly. I'm trying to say that this check should check if `Math.min(range.to + 1, (int) fileSize)` is less than `Integer.Max_VALUE`. Moreover BoundedInoutStream relies on the size of underlying input stream and unless the size of BoundedInputStream is smaller than the size of underlying stream it does not matter how big it is.

jeqo: I see your point, refactoring this validation. Thanks @AnatolyPopov!

ivanyu:
> Not really. At least from the API docs the end position may not be within file bounds:

Gotcha, thanks

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/storage/filesystem/FileSystemStorage.java
```diff
@@ -73,12 +74,16 @@ public InputStream fetch(final String key, final BytesRange range) throws Storag
...
             final Path path = fsRoot.resolve(key);
             final long fileSize = Files.size(path);
-            if (range.to > fileSize) {
-                throw new IllegalArgumentException("position 'to' cannot be higher than the file size, to="
-                    + range.to + ", file size=" + fileSize + " given");
+            if (range.from > fileSize) {
+                throw new InvalidRangeException("Range start position " + range.from
+                    + " is larger than file size " + fileSize);
+            }
+            final long size = Math.min((long) range.to + 1, fileSize); // inclusive end position
```
ivanyu, line 81:
```suggestion
            final long size = Math.min((long) range.to + 1, fileSize); // exclusive end position
```
?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
```diff
@@ -227,12 +231,16 @@ public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegme
                 .fetch(objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.MANIFEST));
 
             final SegmentManifestV1 segmentManifestV1 = mapper.readValue(manifest, SegmentManifestV1.class);
+            final BytesRange range = BytesRange.of(
+                startPosition,
+                Math.min(endPosition, remoteLogSegmentMetadata.segmentSizeInBytes())
```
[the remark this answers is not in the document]
ivanyu: Sorry, ~exclusive~ -> inclusive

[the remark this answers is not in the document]
jeqo: Of course, thank you!

ivanyu (approved):
LGTM.
@AnatolyPopov do you have anything to add?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
```diff
@@ -214,8 +215,11 @@ private void uploadManifest(final RemoteLogSegmentMetadata remoteLogSegmentMetad
     @Override
     public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                                        final int startPosition) throws RemoteStorageException {
-        return this.fetchLogSegment(remoteLogSegmentMetadata, startPosition,
-            remoteLogSegmentMetadata.segmentSizeInBytes());
+        return this.fetchLogSegment(
+            remoteLogSegmentMetadata,
+            startPosition,
+            remoteLogSegmentMetadata.segmentSizeInBytes() - 1
```
AnatolyPopov, line 221: Do we need -1 here? AFAICS we are passing this value further where we do exactly the same, or am I wrong?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/transform/FetchChunkEnumeration.java
```diff
@@ -111,26 +108,26 @@ public InputStream nextElement() {
...
         final boolean isSingleChunk = isAtFirstChunk && isAtLastChunk;
         if (isSingleChunk) {
-            final int toSkip = firstPosition - chunkStartPosition;
+            final int toSkip = range.from - chunkStartPosition;
             try {
                 chunkContent.skip(toSkip);
-                final int chunkSize = lastPosition - firstPosition;
-                chunkContent = new BoundedInputStream(chunkContent, chunkSize);
+                final int chunkSize = range.to - range.from;
+                chunkContent = new BoundedInputStream(chunkContent, chunkSize + 1);
```
AnatolyPopov, line 115: NIt: adding +1 during `chunkSize` calculation on the line above is a little easier to understand.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/transform/FetchChunkEnumeration.java
```diff
@@ -111,26 +108,26 @@ public InputStream nextElement() {
...
                     chunkContent.skip(toSkip);
                 } catch (final IOException e) {
                     throw new RuntimeException(e);
                 }
             }
             if (isAtLastChunk) {
-                final int chunkSize = lastPosition - chunkStartPosition;
-                chunkContent = new BoundedInputStream(chunkContent, chunkSize);
+                final int chunkSize = range.to - chunkStartPosition;
+                chunkContent = new BoundedInputStream(chunkContent, chunkSize + 1);
```
AnatolyPopov, line 130: NIt: adding +1 during `chunkSize` calculation on the line above is a little easier to understand.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
ivanyu: Yes, both will work, but `-1` is technically more correct as the end position is inclusive.

AnatolyPopov (commented):
Some more comments

AnatolyPopov (approved):
LGTM

PR #219, opened 2023-05-11
refactor: introduce Encryption Provider to hide RSA/AES details

jeqo:
Follow up of #190 

Hides AES/RSA details spread across different components.

Replaces #191


jeqo:
Diverged too much, will recreate once/if #217 is adopted

PR #221, opened 2023-05-12
 refactor: Encrypt/Decrypt transforms 

jeqo:
Following up #219 

Refactor encryption/decryption transforms to simplify inputs.

Replaces #192


jeqo:
Diverged too much, will recreate once/if #217 is adopted

PR #222, opened 2023-05-12
Adding in-memory chunk cache implementation

AnatolyPopov:


commons/src/test/java/io/aiven/kafka/tieredstorage/commons/ChunkManagerTest.java
```diff
@@ -0,0 +1,135 @@
...
+@ExtendWith(MockitoExtension.class)
+class ChunkManagerTest  extends AesKeyAwareTest {
+
+    static final byte[] TEST_CHUNK_CONTENT = "0123456789".getBytes();
+    @Mock
+    RemoteLogSegmentMetadata remoteLogSegmentMetadata;
+    @Mock
+    private FileFetcher fileFetcher;
+    @Mock
+    private ObjectKey objectKey;
```
ivanyu, line 62: [Nit] This probably could be just an instance instead of mock

ivanyu (changes requested):
LGTM, but conflicts

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/ChunkManagerTest.java
AnatolyPopov: This is just a little test simplification since we do not really care about generated key here since it's covered in other places.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
```diff
@@ -138,6 +157,17 @@ ObjectStorageFactory objectStorageFactory() {
...
 
+    ChunkCache chunkCache() {
+        final Class<?> chunkCacheClass = getClass(CHUNK_CACHE_CONFIG);
+        if (chunkCacheClass != null) {
+            final ChunkCache chunkCache = Utils.newInstance(chunkCacheClass, ChunkCache.class);
+            chunkCache.configure(this.originalsWithPrefix(CHUNK_CACHE_PREFIX));
+            return chunkCache;
+        } else {
+            return null;
+        }
```
ivanyu, line 168: Consider replacing with `getConfiguredInstance`

ivanyu (changes requested):
LGTM, one small comment

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
```diff
@@ -138,6 +157,10 @@ ObjectStorageFactory objectStorageFactory() {
         return objectFactory;
     }
 
+    ChunkCache chunkCache() {
+        return getConfiguredInstance(CHUNK_CACHE_CONFIG, ChunkCache.class);
```
ivanyu, line 161: Ok, now I realize it wasn't that great a suggestion to use get `getConfiguredInstance`, because it doesn't have a way to cut off the prefix from the names. Sorry for this suggestion :( Let's return to some
```java
        final ChunkCache chunkCache = Utils.newInstance(
            getClass(CHUNK_CACHE_CONFIG), ChunkCache.class
        );
        chunkCache.configure(this.originalsWithPrefix(CHUNK_CACHE_PREFIX));
        return chunkCache;
```

AnatolyPopov: Did a rollback to previous version since null check still makes sense

ivanyu (approved):

PR #223, opened 2023-05-12
Add metrics

ivanyu:
This commit adds only one metric--segment copies per second--but it lays the foundation for further adding more metrics.

### DIFF
--- a/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
+++ b/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManager.java
@@ -24,10 +24,18 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.SequenceInputStream;
 import java.nio.file.Files;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import org.apache.kafka.common.metrics.JmxReporter;
+import org.apache.kafka.common.metrics.KafkaMetricsContext;
+import org.apache.kafka.common.metrics.MetricConfig;
+import org.apache.kafka.common.metrics.Metrics;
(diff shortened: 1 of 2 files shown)


jeqo (commented):
Looks good overall, just a couple of comments regarding implementation:

Yammer version used by Kafka is quite old and kept for compatibility issues: 
- https://issues.apache.org/jira/browse/KAFKA-8721
- https://cwiki.apache.org/confluence/display/KAFKA/KIP-510%3A+Metrics+library+upgrade

Could we state why using this library instead of https://kafka.apache.org/33/javadoc/org/apache/kafka/common/metrics/Metrics.html?

PS. I have bumped [KIP-877](https://cwiki.apache.org/confluence/display/KAFKA/KIP-877%3A+Mechanism+for+plugins+and+connectors+to+register+metrics) thread as it should be the idea way to add plugin specific metrics.

ivanyu:
@jeqo I just didn't think about this. Thank you for pointing this out. Despite I like the old style MBean hierarchy :) I'll switch to the `Metrics`: no need to borrow code in this case.

ivanyu:
Thanks again @jeqo, it turned out simpler and shorter :)

jeqo (approved):
LGTM!

Non-blocking: to follow Metrics usage across Kafka code-base, we may want to have:
- *Metrics: collection of sensors, similar to https://github.com/apache/kafka/blob/f7db6031b84a136ad0e257df722b20faa7c37b8a/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java#L870
- *MetricsRegistry: collection of metric name templates, similar to https://github.com/apache/kafka/blob/01ab888dbd08ccd4b0de9333d21581ce24fe2c3b/clients/src/main/java/org/apache/kafka/clients/producer/internals/SenderMetricsRegistry.java

but we can surely do it later once more metrics are added

PR #225, opened 2023-05-15
Add segment manifest cache

ivanyu:


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param cacheRetention the retention time of items in the cache or {@code null} if infinite retention.
+     */
+    SegmentManifestProvider(final ObjectKey objectKey,
+                            final Long maxCacheSize,
+                            final Duration cacheRetention,
+                            final FileFetcher fileFetcher,
+                            final ObjectMapper mapper) {
+        this.objectKey = objectKey;
+        final var cacheBuilder = Caffeine.newBuilder();
+        if (maxCacheSize != null) {
```
AnatolyPopov, line 50: Am I missing something or instead we could check for -1 here instead of checking twice, first for -1 and then for null?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
```diff
@@ -0,0 +1,89 @@
...
+            cacheBuilder.maximumSize(maxCacheSize);
+        }
+        if (cacheRetention != null) {
```
jeqo, line 53: nit: should we use optional instead?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProviderTest.java
```diff
@@ -0,0 +1,119 @@
...
+            false, null
+        );
+        assertThat(provider.get(REMOTE_LOG_METADATA)).isEqualTo(expectedManifest);
+        assertThat(provider.get(REMOTE_LOG_METADATA)).isEqualTo(expectedManifest);
+        verify(fetcher).fetch(key);
+    }
+
+    @Test
+    void shouldPropagateStorageBackEndException() throws StorageBackEndException {
+        when(fetcher.fetch(anyString()))
```
jeqo, line 113: Same for IOException/JSON related exception?

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProviderTest.java
```diff
@@ -0,0 +1,119 @@
...
+            .thenThrow(new StorageBackEndException("test"));
+        assertThatThrownBy(() -> provider.get(REMOTE_LOG_METADATA))
+            .isInstanceOf(StorageBackEndException.class)
+            .hasMessage("test");
```
jeqo, line 117: Same for IOException/JSON related exceptions?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
AnatolyPopov: I guess optional does not really make sense here, rather same as above - check for -1 here instead of config maybe.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
ivanyu: We're forced to use `-1`: there's no way (that I'm aware of) to explicitly pass `null` from a config file. Kafka uses `-1` as a substitute and I follow suit. However, then I prefer to convert this `-1` to something more proper that conveys special meaning, which is `null` in this case.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProviderTest.java
```diff
@@ -0,0 +1,119 @@
...
+        final String key = "topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.rsm-manifest";
+        when(fetcher.fetch(key))
+            .thenReturn(new ByteArrayInputStream(MANIFEST.getBytes()));
+        final SegmentManifestV1 expectedManifest = new SegmentManifestV1(
+            new FixedSizeChunkIndex(100, 1000, 110, 110),
+            false, null
+        );
+        assertThat(provider.get(REMOTE_LOG_METADATA)).isEqualTo(expectedManifest);
+        assertThat(provider.get(REMOTE_LOG_METADATA)).isEqualTo(expectedManifest);
+        verify(fetcher).fetch(key);
```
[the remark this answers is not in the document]
ivanyu: Good, thanks

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param maxCacheSize the max cache size (in items) or {@code null} if the cache is unbounded.
+     * @param cacheRetention the retention time of items in the cache or {@code null} if infinite retention.
+     */
+    SegmentManifestProvider(final ObjectKey objectKey,
+                            final Long maxCacheSize,
+                            final Duration cacheRetention,
+                            final FileFetcher fileFetcher,
+                            final ObjectMapper mapper) {
+        this.objectKey = objectKey;
+        final var cacheBuilder = Caffeine.newBuilder();
```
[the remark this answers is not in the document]
ivanyu: We don't have an explicit policy (should we?) I generally use `var` to hide cumbersome types that don't matter, like this builder type.

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProviderTest.java
ivanyu: Yes, done this

commons/build.gradle
```diff
@@ -28,4 +28,5 @@ dependencies {
     compileOnly "com.github.luben:zstd-jni:$zstdVersion"
     testImplementation "com.github.luben:zstd-jni:$zstdVersion"
     implementation "commons-io:commons-io:2.11.0"
+    implementation 'com.github.ben-manes.caffeine:caffeine:3.1.4'
```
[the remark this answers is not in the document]
ivanyu: The Kafka implementations are rather primitive and don't seem suitable for highly-concurrent access and fancy eviction policies. Caffeine on the other hand is a mature library with seemingly all the features we may need. @AnatolyPopov is also going to use it for the disk cache anyway, so no new unnecessary dependency.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
```diff
@@ -37,6 +38,20 @@ public class UniversalRemoteStorageManagerConfig extends AbstractConfig {
     private static final String OBJECT_STORAGE_KEY_PREFIX_CONFIG = "key.prefix";
     private static final String OBJECT_STORAGE_KEY_PREFIX_DOC = "The object storage path prefix";
 
+    private static final String SEGMENT_MANIFEST_CACHE_PREFIX = "segment.manifest.cache.";
+    private static final String SEGMENT_MANIFEST_CACHE_SIZE_CONFIG = SEGMENT_MANIFEST_CACHE_PREFIX + "size";
+    private static final Long SEGMENT_MANIFEST_CACHE_SIZE_DEFAULT = 1000L;  // TODO consider a better default
```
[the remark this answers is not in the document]
ivanyu: I thought about that. Technically can be done with weights as you propose.
My thinking is that it may be more convenient from the operation point of view to specify the number of elements instead of bytes. But may be vice versa, depending on the situation. Ideally we should have a possibility for both limits. How do you feel if we make a ticket to address this separately?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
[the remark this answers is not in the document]
jeqo: Sure, make sense. I'm +1 on been relaxed about it as long as we agree on the min jdk features that can be used -- JDK 11 been the minimal version for this project. 

cc @AnatolyPopov as we have touch this in previous reviews.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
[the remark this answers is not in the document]
AnatolyPopov: We can create a separate ticket foe this, yes. But IMO from operational perspective count does not say anything unless you know the chunk size(that can be non constant in case we decide to cache compressed chunks). On the other hand if you can estimate the size of the cache then you can guess appropriate limits plus expose the size as a metric AFAIU. And having such a metric we can create a CLI tool to free up the memory consumed by cache in case of any unexpected situations.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
AnatolyPopov: In the cache implementation -1 is a constant called `UNSET_INT`, maybe we should use that approach to avoid having 2 ifs instead of 1?

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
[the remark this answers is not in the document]
AnatolyPopov: @jeqo as I pointed out last time as far as I remember: I'm not against var's and at the same time do not really like them. We agreed at that time that we perhaps need to have an agreement on this but also IMO sometimes vars make sense, sometimes they decrease the readability and it's hard to invent a generic rule. So I would leave it as it is and unless we feel that var improves the readability I would avoid using it. Here the return type of `newBuilder` is a generic `Caffeine` so you do not need to guess the type when using var, so I think it's fine.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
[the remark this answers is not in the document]
ivanyu: I see your point and it's valid. I will replace the item size limit with weight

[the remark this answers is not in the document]
ivanyu: I managed to calculate object sizes in the run time with [JOL](https://github.com/openjdk/jol). However, I don't feel much like bringing this library as a dependency. It's pretty hacky, uses `Unsafe` and stuff like that and it was probably never meant to run in production.
So, we'll have to live with the item count limit. Considering, we're not speaking about using the last bits of memory for this cache (it should be probably single digit megabytes), this should be fine. We will run some estimate and add to the README how much a typical item will weigh on popular setups.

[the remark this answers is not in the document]
AnatolyPopov: Discussed more offline and I agree, at least for now count is fine.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
```diff
@@ -0,0 +1,89 @@
...
+        this.cache = cacheBuilder.build();
+        this.fileFetcher = fileFetcher;
+        this.mapper = mapper;
+    }
+
+    SegmentManifest get(final RemoteLogSegmentMetadata remoteLogSegmentMetadata)
+        throws StorageBackEndException, IOException {
+        final String key = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.MANIFEST);
+        try {
+            return cache.get(key, k -> {
```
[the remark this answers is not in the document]
ivanyu: Ok, it seems nicely substitutable with the async cache, see the fixup commit.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
ivanyu: You cannot pass `-1`, it'll fail with `maximum size must not be negative`

ivanyu:
Reviewable again

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
AnatolyPopov: Looks like there is a total misunderstanding here. When I was talking about 2 if statements I meant this https://github.com/aiven/tiered-storage-for-apache-kafka/blob/644fadb0dea021240210eaed5a8163bcdb7fad4c/commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java#L194
We are checking for -1 there and returning `null` and then we are checking for `null` here. So my point was that we can pass the value without check from config to here and here have a constant like `UNSET_INT` with the value `-1` to check if the value from config is equal to `UNSET_INT`
I hope now it's clear enough and yes, I'm aware of the fact that -1 can not be passed as an argument. Also the same is probably applicable to `cacheRetention`.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
```diff
@@ -157,6 +189,22 @@ ObjectStorageFactory objectStorageFactory() {
         return objectFactory;
     }
 
+    Long segmentManifestCacheSize() {
```
AnatolyPopov, line 192: Do we really need to make it `Long` and not primitive `long`? I guess it's never null here because of config definition and moreover there is an unboxing on the next line. As I suggested in another comment I would get rid of null-ability here using `long` and returning raw value and check for -1 in the other place.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
ivanyu: Yeah, I got you point.
What I mean is this: At the moment, the special meaning for `-1` exists only in the config definition file. OTOH `null` is universally used for "no value". So I chose here to deal with `-1` no more than necessary and pass `null` around instead.
If we don't do this, at all the places where `maxCacheSize` is used, we have to remember (carry it in some documentation or our heads) that there's some special meaning for `-1`, which we need to check. 3 lines of a trivial `if` is a small price to pay for not having to do this.
Admittedly, this case is simple and the difference is not clear cut. But it's just basic hygiene.

AnatolyPopov: Why not to use Optional if that's the intent? That will clearly say that we need to check for value presence.

ivanyu: Yes, let me switch to `Optional`

ivanyu: Done

AnatolyPopov: Or well, it's clear why not to. But I still do not see a big difference in remembering to check for -1 or for null since -1 has a common lack of value interpretation for primitive arguments like sizes or durations. As for spreading the checks across the code with `maxCacheSize` usage it will anyways take place but with null unless we create a generic wrapper for Caffeine and do it in a single place and I will try to do it some point. Lets not spend more time on this one and as I said I will try to address this later in another PR.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfig.java
AnatolyPopov: Will be addressed later in another PR as mentioned in another comment.

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
AnatolyPopov: Does not seem to be a good idea to have an Optional constructor argument. as(again) mentioned in another thread I will try to address this in a different way a little later.

ivanyu: Well, `Optional` is not recommended as a parameter in general case because of some valid reasons. However, values coming from configuration seem like a pretty valid special case where `Optional`s are just fine.

AnatolyPopov (approved):
LGTM

commons/src/test/java/io/aiven/kafka/tieredstorage/commons/UniversalRemoteStorageManagerConfigTest.java
```diff
@@ -40,6 +42,8 @@ void minimalConfig() {
             )
         );
         assertThat(config.objectStorageFactory()).isInstanceOf(ObjectStorageFactory.class);
+        assertThat(config.segmentManifestCacheSize()).isEqualTo(Optional.of(1000L));
```
AnatolyPopov, line 45: Nit(here and in other places):
```suggestion
        assertThat(config.segmentManifestCacheSize()).hasValue(1000);
```

ivanyu: Great, fixed

ivanyu:
Squashed

jeqo (approved):

commons/src/main/java/io/aiven/kafka/tieredstorage/commons/SegmentManifestProvider.java
```diff
@@ -0,0 +1,116 @@
...
+            final Supplier<SegmentManifest> fetcher = () -> {
+                try (final InputStream is = fileFetcher.fetch(key)) {
+                    return mapper.readValue(is, SegmentManifest.class);
+                } catch (final StorageBackEndException | IOException e) {
+                    // Wrap these exceptions to unwrap them later.
+                    throw new CompletionException(e);
+                }
+            };
+            return CompletableFuture.supplyAsync(fetcher, executor);
+        }
```
ben-manes, line 87: fyi, I think you could have used `CacheLoader.load(key)` and let's default `asyncLoad` wrap that with a `supplyAsync`. A small quality-of-life nit.

ben-manes: I think it would just be the following, assuming might prefer it inlined as a lambda
```java
this.cache = cacheBuilder.buildAsync(key -> {
  try (final InputStream is = fileFetcher.fetch(key)) {
    return mapper.readValue(is, SegmentManifest.class);
  }
}
```

AnatolyPopov: Thanks a lot @ben-manes, looks interesting and I guess will will consider this :+1:

Issue #177, opened 2023-04-20
[New implementation] Consider renaming object.storage.factory setting to something better


ivanyu:

ivanyu:
Yes, this is pretty good. Let's do a rename once we've done with the flow of PRs

jeqo:
Related discussion: https://github.com/aiven/tiered-storage-for-apache-kafka/pull/227

jeqo:
Following up the discussion on #227, I would like to propose:

- Rename `ObjectStorageFactory` to `StorageBackEnd`
- Move `FileFetcher`, `FileUploader`, `FileDeleter` methods to `StorageBackEnd`
- Make `FileSystemStorage` and `S3Storage` to implement `StorageBackEnd` interface
- Move `configure` from `*StorageFactory` to `*Storage`
- Remove `*StorageFactory`
- Remove `File*`
- Rename configuration `object.storage.factory` to `storage.backend.class.name` (following other similar configs in Kafka)
- Rename configuration `storage.*` for Storage configs

cc @AnatolyPopov @ivanyu

ivanyu:
So, effectively the `UniversalRemoteStorageManagerConfig` becomes the factory for storages, if I understand the idea. Sounds good.

I propose to `BackEnd` -> `Backend` as it's more widely used in the industry.

I also propose to have a different instances of the storage for uploading and downloading even after refactoring, so they don't have implicit contention on anything in their completely different use cases.

jeqo:
Sure, I have updated the proposal.

If we want to keep to instances, it would be better to keep two different interfaces. I have reflected that on the proposal as well.

jeqo:
Sure. I'd lean for not group them to have a instance per operation type.

I'm still hesitant if the implicit contention mentioned before needs to be solved at this level, or if already solved by the back-end client. But we can check later if needed when more back-ends are adopted.

See #234

PR #227, opened 2023-05-16
refactor: add base storage factory test

jeqo:
Following up: #210 

Moving end-to-end storage test (upload/fetch/delete) to a base test for storage factory, and leave individual op tests on storage test class.


jeqo:
Thanks @AnatolyPopov 

Fair points. Storage Factory does give access to those interfaces and it's also configurable. Therefore, the goal of this test was to ensure configs are parsed and interfaces accessed by having an end-to-end flow.

Though, I agree Storage Factory doesn't add _much_ and we may consider flatten the storage layers a bit (e.g. have Storage class implement `configurable` and storage interfaces), but this could be discussed in another thread (e.g. #177)

If we agree to discuss StorageFactory somewhere else and tackle this test later, I think we can drop this PR; but if we plan to keep StorageFactory we can redefine the scope of this test, wdyt?

AnatolyPopov:
I think we should reshuffle the code a bit and if we will proceed with #177 then I would maybe bring a little bit more logic there, e.g. implementing those 3 interfaces there for example and getting rid of calls to methods like `fileFetcher()` that are spread through the code base and do not make to much sense to me. And only after that this refactoring IMO will make sense. I would continue this discussion in #177 and first make a decision on that one, e.g. how we should proceed on that one.

AnatolyPopov:
As of current implementation tests are in the places where they should be.

PR #229, opened 2023-05-16
Document compression

ivanyu:


README.md
```diff
@@ -68,15 +68,15 @@ This encoding scheme allows indices to be stored (cached) in memory and accessed
 
 ### Compression
 
-TBD
+Optional data compression with Zstandard is supported. It can be disabled, enabled, or enabled conditionally based on a heuristic.
```
jeqo, line 71: > It can be disabled, enabled, or enabled conditionally based on a heuristic.

Should we describe here some initial defaults? e.g. compression is disabled by default.
Also, from this intro I guess there will be at least 2 flags? e.g. `compression.enabled` and `compression.heuristic` or similar. Or should heuristic be always on? If this is to be defined, let's create an issue to track.

jeqo (commented):
Thanks @ivanyu, looks good to be. just some comments on flags and defaults to keep track.

README.md
ivanyu: I'm not sure that we should put this here. It's the design section and the usage documentation will follow separately.

PR #234, opened 2023-05-23
refactor: define storage back-end

jeqo:
Redefine storage layer interfaces, introducing Storage BackEnd and removing ObjectStorageFactory.

Resolves: #177


commons/src/main/java/io/aiven/kafka/tieredstorage/commons/RemoteStorageManager.java
```diff
@@ -115,7 +119,9 @@ public RemoteStorageManager() {
     public void configure(final Map<String, ?> configs) {
         Objects.requireNonNull(configs, "configs must not be null");
         final RemoteStorageManagerConfig config = new RemoteStorageManagerConfig(configs);
-        objectStorageFactory = config.objectStorageFactory();
+        fetcher = config.storage();
+        uploader = config.storage();
+        deleter = config.storage();
```
ivanyu, line 124: Reviewer note (probably should have been discussed earlier): `AmazonS3Client` is thread-safe and it's OK that we may use these objects concurrently from different threads.

ivanyu (approved):

PR #242, opened 2023-05-26
Stop using Bouncy Castle for AES

ivanyu:
The Bouncy Castle is rather slow, so we decided to switch to the JDK own implementation for AES in both prod code and tests.

`Decryption` and `Encryption` interfaces became slightly different to the point where it's cumbersome trying to keep them, so got rid of them.


jeqo (approved):
LGTM 👍🏽

PR #245, opened 2023-05-31
Adding disk-based chunk cache

AnatolyPopov:


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactory.java
```diff
@@ -0,0 +1,58 @@
...
+                    .getDeclaredConstructor(ChunkManager.class, Executor.class)
+                    .newInstance(chunkManager, executor);
+                cachingChunkManager.configure(config.originalsWithPrefix(ChunkManagerFactoryConfig.CHUNK_CACHE_PREFIX));
+                return cachingChunkManager;
+            } catch (final ReflectiveOperationException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+}
```
jeqo, line 58: We could refactor this factory to have a clearer hierarchy of `ChunkManager`s: 
 
```suggestion
    public CachingChunkManager<?> initChunkManager(final DefaultChunkManager chunkManager,
                                         final Executor executor) {
        try {
            final CachingChunkManager<?> cachingChunkManager = (CachingChunkManager<?>) config
                .cacheType()
                .getChunkManagerClass()
                .getDeclaredConstructor(DefaultChunkManager.class, Executor.class)
                .newInstance(chunkManager, executor);
            cachingChunkManager.configure(config.originalsWithPrefix(ChunkManagerFactoryConfig.CHUNK_CACHE_PREFIX));
            return cachingChunkManager;
        } catch (final ReflectiveOperationException e) {
            throw new RuntimeException(e);
        }
    }
}
```

Then, move the "no caching" condition to the RSM as always needed, and call caching factory conditionally: 

```java
        final DefaultChunkManager defaultChunkManager = new DefaultChunkManager(fetcher, objectKey, aesEncryptionProvider);
        final ChunkManagerFactoryConfig chunkManagerFactoryConfig = new ChunkManagerFactoryConfig(configs);
        if (chunkManagerFactoryConfig.cacheType() == CacheType.NO_CACHE) {
            chunkManager = defaultChunkManager;
        } else {
            final ChunkManagerFactory chunkManagerFactory = new ChunkManagerFactory();
            chunkManagerFactory.configure(configs);
            chunkManager = chunkManagerFactory.initChunkManager(defaultChunkManager, executor);
        }
```

We can then enforce `CachingChunkManager` requires `DefaultChunkManager` as input instead of any implementation, clarifying the wrapping.

```java
    public CachingChunkManager(final DefaultChunkManager chunkManager, final Executor executor) {
        this.chunkManager = chunkManager;
        this.executor = executor;
    }
```

Finally, `ChunkManager` and `DefaultChunkManager` can be moved back to the root directory, and rename the `chunkmanager` directory back to `cache`. `ChunkManagerFactory` can be also renamed to `CachingChunkManagerFactory` or move this initialization to `RSMConfig`.

Hopefully this leaves all the caching details better contained in its own module

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,125 @@
...
+        final ChunkKey chunkKey = new ChunkKey(id, chunkId);
+        final AtomicReference<InputStream> result = new AtomicReference<>();
+        try {
+            return cache.asMap().compute(chunkKey, (key, val) -> CompletableFuture.supplyAsync(() -> {
+                if (val == null) {
+                    try {
+                        final InputStream chunk = chunkManager.getChunk(remoteLogSegmentMetadata, manifest, chunkId);
+                        final T t = this.cacheChunk(chunkKey, chunk);
+                        result.getAndSet(chunkToStream(t));
+                        return t;
```
jeqo, line 69: We may save one transformation when writing to cache -- unless streams are different?:
```suggestion
                        final InputStream chunk = chunkManager.getChunk(remoteLogSegmentMetadata, manifest, chunkId);
                        cacheChunk(chunkKey, chunk);
                        result.getAndSet(chunk);
                        return t;
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,125 @@
...
+        this.chunkManager = chunkManager;
+        this.executor = executor;
+    }
+
+    public InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
+                                final SegmentManifest manifest,
+                                final int chunkId) throws StorageBackendException, IOException {
+        final Uuid id = remoteLogSegmentMetadata.remoteLogSegmentId().id();
+        final ChunkKey chunkKey = new ChunkKey(id, chunkId);
+        final AtomicReference<InputStream> result = new AtomicReference<>();
```
jeqo, line 61: Could we document why we need this atomic reference?

jeqo (commented):
Some initial thoughts

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactory.java
AnatolyPopov: I specifically move these things out of RSM since I believe it should not know anything about ChunkManager implementation and it does not matter if caching is enabled or not before I had it implemented exactly like you suggesting. Also regarding package naming, it's not only about caching but also partially chunking. One idea I had was creating the cache subpackage in chunkmanager package maybe and moving the caching part there.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
AnatolyPopov: We can not, when we are caching the stream is read.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
AnatolyPopov: I wrote some clarification avoiding the reasoning behind AtomicReference since it's not really needed but used for convinience.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
```diff
@@ -126,14 +121,6 @@ public class RemoteStorageManagerConfig extends AbstractConfig {
...
         );
 
-        CONFIG.define(
-            CHUNK_CACHE_CONFIG,
-            ConfigDef.Type.CLASS,
-            null,
-            ConfigDef.Importance.MEDIUM,
-            CHUNK_CACHE_DOC
-        );
-
```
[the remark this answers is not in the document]
AnatolyPopov: The reasoning is rather simple, it's not really possible to instatiate Caffeine without exposing most of the implementation details of different implementations to the config since configuration depends on method implemented in specific subclasses. And factory brings additional convinies by incapsulating implementation details of ChunkManagers into separate place and not polluting configs/RSM with it.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkKey.java
```diff
@@ -33,6 +33,14 @@ public ChunkKey(final Uuid uuid, final int chunkId) {
...
     }
 
+    public Uuid getUuid() {
+        return uuid;
+    }
+
+    public int getChunkId() {
+        return chunkId;
+    }
+
```
jeqo, line 43: On BytesRange and other classes we used public fields, should we replace these getters with public fields?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DefaultChunkManager.java
```diff
@@ -62,7 +55,9 @@ public ChunkManager(final ObjectFetcher fetcher,
     public InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata, final SegmentManifest manifest,
                                 final int chunkId) throws StorageBackendException {
         final Chunk chunk = manifest.chunkIndex().chunks().get(chunkId);
-        final InputStream chunkContent = getChunkContent(remoteLogSegmentMetadata, chunk);
+        final String segmentKey = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.LOG);
+
```
jeqo, line 59: nit: maybe place this new line after the fetcher call to separate the detransform logic

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+import com.github.benmanes.caffeine.cache.AsyncCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalListener;
+import com.github.benmanes.caffeine.cache.Scheduler;
+import com.github.benmanes.caffeine.cache.Weigher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class CachingChunkManager<T> implements Configurable, ChunkManager {
+    private static final Logger log = LoggerFactory.getLogger(CachingChunkManager.class);
```
jeqo, line 45: nit: new line after log to separate static finals with different letter case

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final long GET_TIMEOUT_SEC = 10;
```
jeqo, line 46: nit: new line to separate statics

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+    private final ChunkManager chunkManager;
+    private final Executor executor;
```
jeqo, line 48: nit: new line to separate finals

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+                throw (IOException) e.getCause();
+            }
+
+            throw new RuntimeException(e);
+        } catch (final InterruptedException | TimeoutException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public abstract InputStream chunkToStream(final T cachedChunk);
```
jeqo, line 116: nit: 
```suggestion
    public abstract InputStream cachedChunkToInputStream(final T cachedChunk);
```

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -146,13 +148,12 @@ public void configure(final Map<String, ?> configs) {
...
-            aesEncryptionProvider,
-            config.chunkCache()
-        );
-
+        final ChunkManagerFactory chunkManagerFactory = new ChunkManagerFactory();
+        chunkManagerFactory.configure(configs);
+        chunkManager = chunkManagerFactory.initChunkManager(fetcher,
+                                                            objectKey,
+                                                            aesEncryptionProvider,
+                                                            executor);
```
jeqo, line 156: Not sure if it's an issue, but shouldn't these executors be specific per component?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+                    try {
+                        final T cachedChunk = val.get();
+                        result.getAndSet(chunkToStream(cachedChunk));
+                        return cachedChunk;
+                    } catch (InterruptedException | ExecutionException e) {
+                        throw new CompletionException(e);
+                    }
+                }
+            }, executor)).thenApplyAsync(t -> {
+                log.info("Returning stream");
```
jeqo, line 91: nit: maybe debug?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,133 @@
...
+                return result.get();
+            }).get(GET_TIMEOUT_SEC, TimeUnit.SECONDS);
+        } catch (final ExecutionException e) {
+            // Unwrap previously wrapped exceptions if possible.
+
```
jeqo, line 96: nit: rm line?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManager.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
+
+import io.aiven.kafka.tieredstorage.manifest.SegmentManifest;
+import io.aiven.kafka.tieredstorage.storage.StorageBackendException;
+
+public interface ChunkManager {
+    InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
+                                final SegmentManifest manifest,
+                                final int chunkId) throws StorageBackendException, IOException;
```
jeqo, line 30:
```suggestion
    InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                         final SegmentManifest manifest,
                         final int chunkId) throws StorageBackendException, IOException;
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.kafka.common.Configurable;
+
+import io.aiven.kafka.tieredstorage.ObjectKey;
+import io.aiven.kafka.tieredstorage.security.AesEncryptionProvider;
+import io.aiven.kafka.tieredstorage.storage.ObjectFetcher;
+
+public class ChunkManagerFactory implements Configurable {
+    private ChunkManagerFactoryConfig config;
+
+
```
jeqo, line 31: nit: rm line

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,99 @@
...
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public Path cacheChunk(final ChunkKey chunkKey, final InputStream chunk) throws IOException {
+        final Path tempChunkPath = config.cacheTempPath().resolve(chunkKey.getUuid().toString())
+                .resolve(String.valueOf(chunkKey.getChunkId()));
+        Files.createDirectories(tempChunkPath.getParent());
```
jeqo, line 57: Avoid checking dirs if path already exists
```suggestion
if (!Files.exists(tempChunkPath.getParent())
        Files.createDirectories(tempChunkPath.getParent());
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,99 @@
...
+        final Path tempCached = Files.write(tempChunkPath, IOUtils.toByteArray(chunk));
```
jeqo, line 58: Wondering if readAllBytes is enough?
```suggestion
        final Path tempCached = Files.write(tempChunkPath, chunk.readAllBytes());
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManagerConfig.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package io.aiven.kafka.tieredstorage.chunkmanager;
+
+import java.nio.file.Path;
+import java.util.Map;
+
+import org.apache.kafka.common.config.ConfigDef;
+
+public class DiskBasedCachingChunkManagerConfig extends CachingChunkManagerConfig {
```
jeqo, line 24: Could we describe here why do we need 2 paths?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,99 @@
...
+        log.info("Chunk file has been stored to temporary caching directory {}", tempCached);
+        final Path cachedChunkPath = config.cachePath()
+                .resolve(chunkKey.getUuid().toString())
+                .resolve(String.valueOf(chunkKey.getChunkId()));
+        Files.createDirectories(cachedChunkPath.getParent());
```
jeqo, line 63: same as previous

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,99 @@
...
+        final Path newPath = Files.move(tempCached, cachedChunkPath, ATOMIC_MOVE);
```
jeqo, line 64: Could we document why are we moving files? It may not be obvious why we need this instead of writing once.

jeqo (commented):
Some additional comments

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.github.benmanes.caffeine.cache.RemovalListener;
+import com.github.benmanes.caffeine.cache.Weigher;
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.nio.file.StandardCopyOption.ATOMIC_MOVE;
+
+public class DiskBasedCachingChunkManager extends CachingChunkManager<Path> {
```
jeqo, line 34: non-blocking, but may be worth discussing if we should have an integrity mechanism for data on-disk

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
AnatolyPopov: So far the decision was made to use single executor. But we can reconsider.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
AnatolyPopov: Keeping the original style since this is a copy+paste, will address formatting in both places together in a separate PR.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManager.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactory.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManagerConfig.java
AnatolyPopov: Was discussed before and was going to address this later but done already.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
```diff
@@ -0,0 +1,111 @@
...
+import com.github.benmanes.caffeine.cache.Weigher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.nio.file.StandardCopyOption.ATOMIC_MOVE;
+
+public class DiskBasedCachingChunkManager extends CachingChunkManager<Path> {
+
+    private static final Logger log = LoggerFactory.getLogger(DiskBasedCachingChunkManager.class);
+    static final String TEMP_CACHE_DIRECTORY = "temp";
```
jeqo, line 36: nit: `TMP` instead of `TEMP` seem to be more broadly used.
```suggestion
    static final String CACHE_TMP_DIRECTORY = "tmp";
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactory.java
jeqo: Make sense. Agree we can look into `chunkmanager` refactoring later if we cannot think of a better organization atm.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DefaultChunkManager.java
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
AnatolyPopov: This is also not possible to enforce by tooling so I would leave it as is.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
AnatolyPopov: Added some javadocs to the method.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
AnatolyPopov: There was a discussion regarding verification of hash sum or something similar but this will probably addressed later.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
jeqo: Reconsidering this, as `temp` already used in our code base.

jeqo (dismissed):
LGTM, thanks @AnatolyPopov !

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DiskBasedCachingChunkManager.java
jeqo: Looking into Java options when writing, wouldn't `SYNC` [1] help us to avoid writing 2 files?

[1] https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#SYNC

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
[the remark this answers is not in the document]
ivanyu: I could be possible to make it configurable with classes if we refactor this a bit, though. For example, create `AbstractChunkManager` that would every other inherit from (maybe indirectly) and move `getChunk` logic there. Thus, we can get rid of wrapping and all the non-abstract classes will have the uniform constructor with `final ObjectFetcher fetcher, final ObjectKey objectKey, final AesEncryptionProvider aesEncryptionProvider` args.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/CachingChunkManager.java
```diff
@@ -0,0 +1,132 @@
...
+                                final int chunkId) throws StorageBackendException, IOException {
+        final Uuid id = remoteLogSegmentMetadata.remoteLogSegmentId().id();
+        final ChunkKey chunkKey = new ChunkKey(id, chunkId);
+        final AtomicReference<InputStream> result = new AtomicReference<>();
+        try {
+            return cache.asMap().compute(chunkKey, (key, val) -> CompletableFuture.supplyAsync(() -> {
+                if (val == null) {
+                    try {
+                        final InputStream chunk = chunkManager.getChunk(remoteLogSegmentMetadata, manifest, chunkId);
+                        final T t = this.cacheChunk(chunkKey, chunk);
```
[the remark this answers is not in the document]
AnatolyPopov: Maybe I'm missing something also but is not it so that the actual reading from S3 will happen after the addition of the CompletableFuture into the cache? Inside `compute` method in `ConcurrentHashMap` the remapping function will be called  but it just creates the instance of CompletableFuture that will be run async because of `supplyAsync` call.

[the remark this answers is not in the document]
ivanyu: Yep, you're right. Sorry, my bad.

jeqo (approved):

core/src/test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerMetricsTest.java
```diff
@@ -83,7 +85,10 @@ static void setup(@TempDir final Path tmpDir,
             "chunk.size", "123",
             "storage.backend.class",
             "io.aiven.kafka.tieredstorage.storage.filesystem.FileSystemStorage",
-            "storage.root", target.toString()
+            "storage.root", target.toString(),
+            "chunk.cache.path", tmpDir.resolve("cache").toString(),
+            "chunk.cache.temp.path", tmpDir.resolve("tempCache").toString(),
```
ivanyu, line 90: This setting now doesn't exists AFAIU

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManager.java
```diff
@@ -0,0 +1,32 @@
...
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
+
+import io.aiven.kafka.tieredstorage.manifest.SegmentManifest;
+import io.aiven.kafka.tieredstorage.storage.StorageBackendException;
+
+public interface ChunkManager {
+
+    InputStream getChunk(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
+                                         final SegmentManifest manifest,
+                                         final int chunkId) throws StorageBackendException, IOException;
```
ivanyu, line 31: [Nit] Indentation

core/src/test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerMetricsTest.java
AnatolyPopov: Yeah, forgot to get rid of it.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManager.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/InMemoryChunkCache.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public byte[] cacheChunk(final ChunkKey chunkKey, final InputStream chunk) throws IOException {
+        try (chunk) {
+            return chunk.readAllBytes();
+        }
+    }
+
+    @Override
+    public RemovalListener<ChunkKey, byte[]> removalListener() {
+        return (key, content, cause) -> log.info("Deleted cached value for key {} from cache."
```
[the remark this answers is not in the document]
AnatolyPopov: Yeah, lets make it debug.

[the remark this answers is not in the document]
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final String CACHE_RETENTION_CONFIG = "retention.ms";
+    private static final String CACHE_RETENTION_DOC = "Cache retention time ms";
+    private static final long DEFAULT_CACHE_SIZE_BYTES = 100 * 1024 * 1024;
+    private static final long DEFAULT_CACHE_RETENTION_MS = 600_000;
+
+    private static ConfigDef addCacheConfigs(final ConfigDef configDef) {
+        configDef.define(
+                CACHE_SIZE_CONFIG,
+                ConfigDef.Type.LONG,
+                DEFAULT_CACHE_SIZE_BYTES,
```
ivanyu, line 38: Maybe let's not introduce the default value? Sounds important enough to make users make explicit decision

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public RemovalListener<ChunkKey, Path> removalListener() {
+        return (key, path, cause) -> {
+            try {
+                Files.delete(path);
+            } catch (final IOException e) {
+                log.error("Failed to delete cached file for key {} with path {} from cache directory."
+                              + " The reason of the deletion is {}", key, path, cause, e);
+            }
+            log.info("Deleted cached file for key {} with path {} from cache directory."
```
ivanyu, line 98: Same story with debug, maybe?

ivanyu, line 98: In case of error, the logging is incorrect. I think the success logging should be inside the `try`.

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
```diff
@@ -401,15 +404,20 @@ void testRequiresCompression(final CompressionType compressionType, final boolea
...
+        final Map<String, ?> config = new HashMap<>() {{
+                put("chunk.size", Integer.toString(chunkSize));
+                put("storage.backend.class",
+                    "io.aiven.kafka.tieredstorage.storage.filesystem.FileSystemStorage");
+                put("key.prefix", "test/");
+                put("storage.root", targetDir.toString());
+                put("compression.enabled", "true");
+                put("compression.heuristic.enabled", "true");
+                put("chunk.cache.size", 10000);
+                put("chunk.cache.path", tmpDir.toString());
```
ivanyu, line 416: [Nit] This is unnecessary since we're using the in-mem cache

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+
+public class ChunkCacheConfig extends AbstractConfig {
+    private static final String CACHE_SIZE_CONFIG = "size";
+    private static final String CACHE_SIZE_DOC = "Cache size in bytes";
+    private static final String CACHE_RETENTION_CONFIG = "retention.ms";
+    private static final String CACHE_RETENTION_DOC = "Cache retention time ms";
```
ivanyu, line 30: I think in these two, `-1` should be documented

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Fixed

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
AnatolyPopov: Fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public Path cacheChunk(final ChunkKey chunkKey, final InputStream chunk) throws IOException {
+        final Path tempChunkPath = config.baseCachePath().resolve(TEMP_CACHE_DIRECTORY)
+                .resolve(chunkKey.uuid.toString())
+                .resolve(String.valueOf(chunkKey.chunkId));
+        if (Files.notExists(tempChunkPath.getParent())) {
+            Files.createDirectories(tempChunkPath.getParent());
+        }
+        final Path tempCached = writeToDisk(chunk, tempChunkPath);
+        log.info("Chunk file has been stored to temporary caching directory {}", tempCached);
```
ivanyu, line 71:
```suggestion
        log.debug("Chunk file has been stored to temporary caching directory {}", tempCached);
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+        final Path cachedChunkPath = config.baseCachePath()
+                .resolve(CACHE_DIRECTORY).resolve(chunkKey.uuid.toString())
+                .resolve(String.valueOf(chunkKey.chunkId));
```
ivanyu, line 74: Same question about flatness

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+        if (Files.notExists(cachedChunkPath.getParent())) {
+            Files.createDirectories(cachedChunkPath.getParent());
+        }
+        final Path newPath = Files.move(tempCached, cachedChunkPath, ATOMIC_MOVE);
```
ivanyu, line 78: Maybe if `move` fails, we should delete the temp file?

ivanyu:
We currently don't have any recovery strategy for the disk-based cache. What happens if a process is restarted? I guess, cache files leak. As the most simple recovery strategy, we could unconditionally clean the cache directories when `DiskBasedChunkCache` is instantiated. Sounds reasonable?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+                              + " The reason of the deletion is {}", key, path, cause, e);
+            }
+        };
+    }
+
+    @Override
+    public Weigher<ChunkKey, Path> weigher() {
+        return (key, value) -> {
+            try {
+                return Long.valueOf(Files.size(value)).intValue();
```
ivanyu, line 107: [Nit] I guess that's the same
```suggestion
                return (int) Files.size(value));
```

ivanyu: Also, this is not safe regarding sizes over `Integer.MAX_VALUE`. Should we cap the weight to `Integer.MAX_VALUE`? (Despite this shouldn't ever happen, but just in case)

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+     * Writes the chunk to specified directory on disk.
+     * To be able to understand if the files are fully written to disk during a potential recovery process,
+     * this implementation initially writes them to temporary directory and then atomically copies them
+     * to the actual caching directory.
+     */
+    @Override
+    public Path cacheChunk(final ChunkKey chunkKey, final InputStream chunk) throws IOException {
+        final Path tempChunkPath = config.baseCachePath().resolve(TEMP_CACHE_DIRECTORY)
+                .resolve(chunkKey.uuid.toString())
+                .resolve(String.valueOf(chunkKey.chunkId));
```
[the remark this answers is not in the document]
AnatolyPopov: You mean totally flat inside cache directory or having subdirectories like `<uuid>-<chunk_id>`?

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: For now I think we can do that but at the same time if the request will be retied by broker we will download the chunk again and this can be an infinite loop AFAIU. So maybe we should check maybe if we have the chunk in temp directory already before download? But we can address this later.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: How would you suggest to cap it? The only way that comes to my mind is throwing an exception that the chunk is too big. The purpose `Long.valueOf(Files.size(value)).inValue()` was to get rid of the warning about unsafe conversion from long to int and offload this to JDK despite the fact that this is exactly the same. But you might be right that it's better to just cast.

AnatolyPopov:
> We currently don't have any recovery strategy for the disk-based cache. What happens if a process is restarted? I guess, cache files leak. As the most simple recovery strategy, we could unconditionally clean the cache directories when DiskBasedChunkCache is instantiated. Sounds reasonable?
You mean until we implement the recovery?

AnatolyPopov:
> We currently don't have any recovery strategy for the disk-based cache. What happens if a process is restarted? I guess, cache files leak. As the most simple recovery strategy, we could unconditionally clean the cache directories when DiskBasedChunkCache is instantiated. Sounds reasonable?
You mean until we implement the recovery?

ivanyu:
Yes until then. It's a recovery strategy by itself, maybe we'll live just fine with it

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
ivanyu: > How would you suggest to cap it? The only way that comes to my mind is throwing an exception that the chunk is too big.

Maybe just return max_int if the value >= max_int (compared in the long space). It's a safety net for a not-sure-if-ever-happens case

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: That's what I was thinking about also. AFAIR there was some discussion about implementing some integrity checks. So let's postpone this and just delete for now I guess

ivanyu: Yeah, that's reasonable

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: I will add a warn log message just in case. Otherwise it will be really hard to track if something like this will go wrong.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,119 @@
...
+            return Files.newInputStream(cachedChunk);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Writes the chunk to specified directory on disk.
+     * To be able to understand if the files are fully written to disk during a potential recovery process,
+     * this implementation initially writes them to temporary directory and then atomically copies them
```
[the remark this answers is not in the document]
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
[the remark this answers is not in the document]
AnatolyPopov: Yeah, changed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: FIxed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Done

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Fixed

ivanyu: Yeah, looks good!

Maybe `<=` would work as well.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
```diff
@@ -0,0 +1,133 @@
...
+        if (Files.notExists(cachedChunkPath.getParent())) {
+            Files.createDirectories(cachedChunkPath.getParent());
+        }
+        try {
+            final Path newPath = Files.move(tempCached, cachedChunkPath, ATOMIC_MOVE);
+            log.debug("Chunk file has been moved to cache directory {}", newPath);
+            return newPath;
+        } finally {
+            // Cleaning up the temp directory in case of exceptions during move.
+            Files.deleteIfExists(tempCached);
```
ivanyu, line 84: Maybe some error logging here would be good, too

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Yeah, fixed

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCache.java
AnatolyPopov: Added

AnatolyPopov:
Also added deletion of leftover cached files during startup.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
```diff
@@ -0,0 +1,83 @@
...
+    public DiskBasedChunkCacheConfig(final Map<String, ?> props) {
+        super(configDef(), props);
+        // Cleaning the cache directory since there is no way so far
+        // to reuse previously cached files after broker restart.
+        resetCacheDirectory();
+    }
+
+    private void resetCacheDirectory() {
+        if (Files.exists(baseCachePath())) {
+            try (final var files = Files.walk(baseCachePath());) {
```
ivanyu, line 58: We have `commons-io` in the deps, so can be just `FileUtils.deleteDirectory`

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
```diff
@@ -0,0 +1,83 @@
...
+                final var pathList = files.sorted(reverseOrder()).collect(Collectors.toList());
+                for (final Path path : pathList) {
+                    Files.delete(path);
+                }
+                Files.createDirectories(cachePath());
+                Files.createDirectories(tempCachePath());
```
ivanyu, line 64: This should probably be outside of this `try` on `Files.walk`

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public DiskBasedChunkCacheConfig(final Map<String, ?> props) {
+        super(configDef(), props);
+        // Cleaning the cache directory since there is no way so far
+        // to reuse previously cached files after broker restart.
+        resetCacheDirectory();
+    }
+
+    private void resetCacheDirectory() {
```
ivanyu, line 56: Not sure if it should be in the config class, but we can decide this separately

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
AnatolyPopov: Maybe I'm missing something but does it matter? I don't really want to add one more try-catch..

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
AnatolyPopov: I was hesitating a lot and was moving it back and forth 3 times already :sweat_smile: And this seems to be the cleanest apart from the fact that it's a config class. So idk.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
AnatolyPopov: Well, @jeqo was advocating a lot about using java.nio instead so I thought about avoiding commons-io but here it actually really makes sense. Changing

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfig.java
ivanyu: Since there's no `Files.walk` with its `try`, it looks fine to me

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheConfigTest.java
```diff
@@ -0,0 +1,108 @@
...
+                Map.of(
+                        "size", "-1",
+                        "path", emptyCachePath
+                )
+        );
+        assertThat(diskBasedCachingChunkManagerConfig.baseCachePath()).isEqualTo(Path.of(emptyCachePath));
+    }
+
+    @Test
+    void noneEmptyCachePath() throws IOException {
```
ivanyu, line 69: I guess
```suggestion
    void nonEmptyCachePath() throws IOException {
```
was meant

ivanyu (approved):
LGTM.
One super-minor thing, but we can resolve it later

ivanyu:
Thank you @AnatolyPopov for developing and @jeqo for reviews! 🌄

PR #261, opened 2023-06-06
Add some metrics

ivanyu:
- `segment-copy-time-avg`
- `segment-copy-time-max`
- `segment-fetch-rate`

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -36,6 +36,8 @@ import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
 import org.apache.kafka.common.metrics.Metrics;
 import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.common.metrics.stats.Avg;
+import org.apache.kafka.common.metrics.stats.Max;
 import org.apache.kafka.common.metrics.stats.Rate;
 import org.apache.kafka.common.utils.ByteBufferInputStream;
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -105,14 +113,29 @@ public RemoteStorageManager() {
...
+        segmentCopyTime.add(metrics.metricName("segment-copy-time-max", metricGroup), new Max());
+
+        segmentFetchPerSec = metrics.sensor("segment-fetch");
+        segmentFetchPerSec.add(metrics.metricName("segment-fetch-rate", metricGroup), new Rate());
+
+        segmentFetchToFirstByteTime = metrics.sensor("segment-fetch-to-first-byte-time");
+        segmentFetchToFirstByteTime.add(metrics.metricName("segment-fetch-to-first-byte-time-avg", metricGroup),
+            new Avg());
+        segmentFetchToFirstByteTime.add(metrics.metricName("segment-fetch-to-first-byte-time-max", metricGroup),
+            new Max());
```
jeqo, line 138: Could we add some comments (not sure if here or in README maybe?) about how this metric is going to be used? Others are pretty standard, but time to first byte seems to deliver some insights that are not obvious.

jeqo (commented):
Looks pretty good, just a comment about time to first byte metric.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
ivanyu: I removed this one for now. Need more thinking of how to better implement it and next time I'll add a description

jeqo (approved):
LGTM

PR #266, opened 2023-06-07
Close all source input streams when fetching

ivanyu:


ivanyu:
This apparently fixed non-cached fetching

core/src/main/java/io/aiven/kafka/tieredstorage/transform/FetchChunkInputStream.java
jeqo, line 1: The only issue I find with this approach is that `(Transform/Detransfor)Finisher`s are using the `new SequenceInputStream(finisher)`. 
Should these sequenceInputStream creation be consistent between these enumerations?
Maybe we can add `toSequence()` to these implementations to generates a `SequenceInputStream` out of the enumeration to align without need of a new class.

ivanyu: Yes, I can do this. Maybe as a follow-up?

jeqo (approved):

PR #267, opened 2023-06-07
Rename storage.backend.class.name -> storage.backend.class

ivanyu:
Following the convention from Kafka.


AnatolyPopov (approved):
LGTM

jeqo:
We were following it, at least similar to https://github.com/apache/kafka/blob/9cfc4b9373bf887ec2fb95bc607f4ebd1c8613c0/storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManagerConfig.java#L60 or https://kafka.apache.org/documentation/#brokerconfigs_alter.config.policy.class.name
Though, I can find some inconsistencies, e.g. https://kafka.apache.org/documentation/#consumerconfigs_sasl.login.callback.handler.class

I don't feel strongly about it, but be aligned with remote storage configs sounds ok to me.

ivanyu:
Oops, it seems we've just looked at different parts of Kafka and there indeed were some inconsistency

PR #271, opened 2023-06-08
feat(storage:s3): add s3 multi-part upload part size config

jeqo:
Adds a configuration for multi-part upload. Ideally this part size should be the same as chunk size, so we gain some performance when reading[3], but this should be considered later once we have more experience with different chunk sizes.

Resolves: #233

References:

[1] Overview: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html
[2] Multi-part upload limits: https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
[3] https://docs.aws.amazon.com/whitepapers/latest/s3-optimizing-performance-best-practices/use-byte-range-fetches.html
> Typical sizes for byte-range requests are 8 MB or 16 MB. If objects are PUT using a multipart upload, it’s a good practice to GET them in the same part sizes (or at least aligned to part boundaries) for best performance. GET requests can directly address individual parts; for example, GET ?partNumber=N.

### DIFF
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
@@ -48,6 +48,15 @@ public class S3StorageConfig extends AbstractConfig {
         + "By default, empty value means S3 library will auto-detect. "
         + "Amazon S3 uses virtual hosts by default (true), but other S3-compatible backends may differ (e.g. minio).";
 
+    private static final String S3_MULTIPART_UPLOAD_PART_SIZE_CONFIG = "s3.multipart.upload.part.size";
+    // AWS limits to 5GiB, but 2GiB are used here as ByteBuffer allocation is based on int
+    private static final String S3_MULTIPART_UPLOAD_PART_SIZE_DOC = "Size of parts in bytes to use when uploading. "
+        + "All parts but the last one will have this size. "
+        + "Valid values: between 5MiB and 2GiB";
+    static final int S3_MULTIPART_UPLOAD_PART_SIZE_MIN = 5 * 1024 * 1024; // 5MiB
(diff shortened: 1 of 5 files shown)


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -48,6 +48,14 @@ public class S3StorageConfig extends AbstractConfig {
         + "By default, empty value means S3 library will auto-detect. "
         + "Amazon S3 uses virtual hosts by default (true), but other S3-compatible backends may differ (e.g. minio).";
 
+    private static final String S3_UPLOAD_MULTI_PART_SIZE_BYTES_CONFIG = "s3.upload.multi.part.size.bytes";
```
ivanyu, line 51: Maybe just
```suggestion
    private static final String S3_UPLOAD_MULTI_PART_SIZE_BYTES_CONFIG = "s3.upload.part.size.bytes";
```
?

Also I think for consistency with `chunk.size` we should drop `bytes`.

ivanyu (changes requested):
LGTM, only one concern about the property name

ivanyu (approved):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
AnatolyPopov: Can this lead to a confusion between chunk and part? IMO having multipart in the name is a little easier to understand but I'm fine with both.

ivanyu: You're right, good point. `s3.miltipart.upload.part.size` maybe then?

AnatolyPopov (approved):
LGTM

Issue #125, opened 2023-03-14
the multipart upload performance is not ideal


HenryCaiHaiying:
From the code, it doesn't looks to me it's really using S3 multi-threading for multipart upload:

1. A while loop in S3ClientWrapper#uploadLogFile to break the original segment file into multiple parts and upload them one by one through the custom S3OutputStream: https://github.com/aiven/tiered-storage-for-apache-kafka/blob/main/s3/src/main/java/io/aiven/kafka/tiered/storage/s3/S3ClientWrapper.java#L179
2. In S3OutputStream, the code will try to use S3's multipart upload API to upload the file in multiple chunks.   But since the caller (S3ClientWrapper) already configured the file size to be config.s3StorageUploadPartSize, this would just end up with 1 upload in S3OutputStream

S3s multipart upload is supposed to use multiple threads to upload a big file concurrently onto S3.   The current code path doesn't seem using S3's multipart upload threading.

PR #272, opened 2023-06-09
feat(storage:s3): multi-part upload: upload parts concurrently

jeqo:
Uses `CompletableFuture` to run upload parts requests concurrently, and improve upload performance.

Resolves: #125


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -122,31 +149,32 @@ public void close() throws IOException {
...
+                    .withBucketName(bucketName)
+                    .withKey(key)
+                    .withUploadId(uploadId)
+                    .withPartSize(actualPartSize)
+                    .withPartNumber(partUploads.size() + 1)
+                    .withInputStream(new ByteArrayInputStream(array));
+
+            // Run request async
+            final CompletableFuture<Void> upload =
+                CompletableFuture.supplyAsync(() -> client.uploadPart(uploadPartRequest))
```
AnatolyPopov, line 168: AS we discussed offline already a little bit cleaner approach to aggregate the results as `PartETag`'s using ComplrtableFuture.thenCompose` and getting rid of synchronized list. And instead of relying on the list size when determining a part number a plain integer counter might work. This will simplify the usage of `CompletableFuture` that we have now in 2 places.

AnatolyPopov (changes requested):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jeqo: Sure, was actually testing some approaches. I ended up leaning towards `CompletableFuture.thenCombine` instead. Let me know how it looks.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -122,31 +142,31 @@ public void close() throws IOException {
...
+                .withUploadId(uploadId)
+                .withPartSize(actualPartSize)
+                .withPartNumber(partNumber.incrementAndGet())
+                .withInputStream(new ByteArrayInputStream(array));
+
+            // Run request async
+            partUploads = partUploads.thenCombine(
+                CompletableFuture.supplyAsync(() -> client.uploadPart(uploadPartRequest)),
+                (partETags, result) -> {
+                    partETags.put(result.getPartETag().getPartNumber(), result.getPartETag().getETag());
```
AnatolyPopov, line 162: I was thinking a little further and because of this PartETag reconstructing can we maybe use ConcurrentLinkedQueue here? Then when passing the results further to AWS client we can just use ArrayList constructor that is able to accept this queue as an argument

AnatolyPopov (commented):
Some more thoughts

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jeqo: Good idea, I like it. Applying the suggestion

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -122,31 +139,31 @@ public void close() throws IOException {
...
+                .withInputStream(new ByteArrayInputStream(array));
+
+            // Run request async
+            partUploads = partUploads.thenCombine(
+                CompletableFuture.supplyAsync(() -> client.uploadPart(uploadPartRequest)),
+                (partETags, result) -> {
+                    partETags.add(result.getPartETag());
+                    return partETags;
+                });
         } catch (final Exception e) {
```
AnatolyPopov, line 162: RuntimeException maybe?

AnatolyPopov: Or do we need this catch here at all? At least if I remove the catch tests are passing and the exception I believe will be thrown on call to `get`

jeqo: I'm failing to see how this would help giving that RuntimeException is a subclass of Exception; could you elaborate?

jeqo: Agree. And we the fixes on #282 the transaction should be aborted properly.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -135,102 +136,116 @@ void writesOneByte() throws Exception {
...
     void writesMultipleMessages() throws Exception {
         final int bufferSize = 10;
-        final byte[] message = new byte[bufferSize];
 
         when(mockedS3.initiateMultipartUpload(any(InitiateMultipartUploadRequest.class)))
             .thenReturn(newInitiateMultipartUploadResult());
+
+        final Map<Integer, UploadPartRequest> uploadPartRequests = new ConcurrentHashMap<>();
+        final Map<Integer, byte[]> uploadPartContents = new ConcurrentHashMap<>();
         when(mockedS3.uploadPart(uploadPartRequestCaptor.capture()))
```
AnatolyPopov, line 161: This is weird. What are you mocking? Do we expect this method to answer when we called it with a captor? What will happen it will not be called and then we will try to get a captured value?

AnatolyPopov: And same for other places where captor is used

AnatolyPopov (changes requested):

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
jeqo: Good catch, probably copy-pasted before. But you're right, captured values are not used on this cases and I'm replacing with `any(...)` instead

ivanyu (changes requested):
This approach has a drawback: we can't realistically use big part sizes. We need to target some 100 MB part sizes. With uncontrollable parallelism + memory arrays, this is not going to end well. Mainly we need to get rid of memory arrays and stream data directly from disk.

Ideally we should have something like this:
1. Split the file into parts, virtually, just ranges.
2. Upload parts in parallel from file-based InputStream (applying transfomations, of course).
3. Control parallelism, we need an explicit configurable number.
4. Recombine the chunk index in the end, should be a simple arithmetics.
5. Some mechanism for cancelling all parallel uploads if one of them fails (maybe we don't need this if S3 starts rejecting uploads promptly enough after the multipart upload has been cancelled).

This will lead to changing some interfaces + making it required that the chunks size is a multiple of the chunk size.

PR #274, opened 2023-06-13
Introducing integration tests

AnatolyPopov:


.github/workflows/main_push_workflow.yml
```diff
@@ -22,4 +22,4 @@ jobs:
           java-version: ${{ matrix.java-version }}
           distribution: temurin
       - name: Build with Gradle
-        run: ./gradlew build releaseTarGz
+        run: ./gradlew build integrationTest releaseTarGz
```
jeqo, line 25: Wouldn't `check` be enough?
```suggestion
        run: ./gradlew check releaseTarGz
```

AnatolyPopov: check does not run integration tests unfortunately

jeqo: You're right! I always get confused by the docs https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests but it adds it explicitly to the check with `dependsOn`. I guess we don't see a need to add it to the check lifecycle

jeqo (dismissed):

.github/workflows/main_push_workflow.yml
AnatolyPopov: Also talking about this specific place I don't think releaseTarGz is needed here since we are not publishing the result anywhere

jeqo: I think it was added to exercise the packaging as it's used on the Docker image (maybe we should build the image as part of the workflow instead?) cc @ivanyu

ivanyu:
What would be the distinctive characteristic of integration tests in this project? To me it seems that `RemoteStorageManagerTest` and `S3StorageTest` aren't integration tests, but future tests where we run a broker with tiered storage will be.

Makefile
```diff
@@ -21,7 +21,7 @@ clean:
 	./gradlew clean
 
 build/distributions/tiered-storage-for-apache-kafka-$(VERSION).tgz:
-	./gradlew check releaseTarGz
+	./gradlew build releaseTarGz
```
ivanyu, line 24: Probably running at least unit tests would still make sense for building the Docker image

AnatolyPopov: Build depends on check, check depends on both types of tests.

.github/workflows/main_push_workflow.yml
AnatolyPopov: We perhaps should, changed it back to build + releaseTarGz since build runs all the tests now. Will figure out releaseTarGz necessity and image building later.

jeqo: sounds good to me

ivanyu (approved):
I don't really like the idea of marking `RemoteStorageManagerTest` an integration test, but anyway LGTM

PR #262, opened 2023-06-06
chore: update Docker to new implementation

jeqo:
Update docker to latest implementation and unify with https://github.com/ivanyu/kafka-tiered-storage-demo 

Resolves #224

PR #275, opened 2023-06-13
Gradle files cleanup and version bump

AnatolyPopov:


AnatolyPopov:
@jeqo Fixed + fixed one more place in Dockerfile

docker/kafka/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN git checkout ${AK_BRANCH}
 
 ARG SCALA_VERSION=2.12
 
-RUN ./gradlew -PscalaVersion=${SCALA_VERSION} testJar releaseTarGz
+RUN ./gradlew -PscalaVersion=${SCALA_VERSION} testJar distTar
```
jeqo, line 22: I think this one is not required as it's building upstream kafka, not the plugin

jeqo: We will get rid of this file anyway on #262

AnatolyPopov: Ah, well, you right, was doing just grep + replace

AnatolyPopov: Fixed

docker/kafka/Dockerfile
```diff
@@ -37,4 +37,4 @@ COPY --from=base /kafka/storage/api/build/libs/kafka-storage-api-${KAFKA_VERSION
 ENV KAFKA_HOME /kafka
 WORKDIR ${KAFKA_HOME}
 ENV PATH $PATH:${KAFKA_HOME}/bin
-CMD [ "kafka-server-start.sh" ]
\ No newline at end of file
+CMD [ "kafka-server-start.sh" ]
```
jeqo: nit: just wondering if this is intentional or sometimes github get's confused with new lines at EOF.

AnatolyPopov: Not intentional but new lines at the end are required in most of our projects and when touching file having IDE configured to add them this happens sometimes :) I would prefer to keep it if you don't mind, despite the fact that it's unrelated to PR.

build.gradle
```diff
@@ -118,16 +116,22 @@ subprojects {
...
+            from(project(":storage:core").jar)
+            from(project(":storage:core").configurations.runtimeClasspath)
+            from(project(":storage:s3").jar)
+            from(project(":storage:s3").configurations.runtimeClasspath)
+            from(project(":storage:filesystem").jar)
+            from(project(":storage:filesystem").configurations.runtimeClasspath)
+        }
+    }
+}
+
```
ivanyu, line 133: Checked, this seems to produce identical content with the previous `releaseTarGz`.

ivanyu (approved):

build.gradle
AnatolyPopov: Exactly, did the same while was refactoring.

docker/kafka/Dockerfile
AnatolyPopov: Hm, weird.

PR #279, opened 2023-06-14
chore(build): disable javadoc lint

jeqo:
Javadoc docs [1] :
```
-quiet
    Shuts off messages so that only the warnings and errors appear to make them easier to view. It also suppresses the version string. 
```

Doclet[2]:

```
-Xdoclint all,-group: Enables all checks except group checks
```

[1]: https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html
[2]: https://docs.oracle.com/en/java/javase/17/docs/specs/man/javadoc.html#additional-options-provided-by-the-standard-doclet

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -173,6 +173,11 @@ subprojects {
         dependsOn tasks.named("integrationTest")
     }
 
+    tasks.javadoc {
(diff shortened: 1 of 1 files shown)


AnatolyPopov (approved):
LGTM

PR #280, opened 2023-06-14
feat: measure object upload bytes

jeqo:
Changes:
- To enable metrics to capture the number of bytes uploaded to storage backends (e.g. after transforms), `upload` is expanded to return the number of bytes uploaded.
- Metrics to measure object uploads per suffix type are added.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -252,7 +252,12 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
         throws IOException, StorageBackendException {
         final String fileKey = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.LOG);
         try (final var sis = transformFinisher.toInputStream()) {
-            uploader.upload(sis, fileKey);
+            final var bytes = uploader.upload(sis, fileKey);
+            metrics.recordObjectUpload(
+                remoteLogSegmentMetadata.remoteLogSegmentId().topicIdPartition().topicPartition(),
+                ObjectKey.Suffix.LOG,
+                bytes
+            );
         }
     }
 
@@ -260,9 +265,15 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
                                  final InputStream index,
                                  final IndexType indexType)
         throws StorageBackendException, IOException {
-        final String key = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.fromIndexType(indexType));
+        final var suffix = ObjectKey.Suffix.fromIndexType(indexType);
+        final String key = objectKey.key(remoteLogSegmentMetadata, suffix);
         try (index) {
-            uploader.upload(index, key);
+            final var bytes = uploader.upload(index, key);
+            metrics.recordObjectUpload(
+                remoteLogSegmentMetadata.remoteLogSegmentId().topicIdPartition().topicPartition(),
+                suffix,
+                bytes
+            );
         }
     }
 
@@ -273,7 +284,12 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
         final String manifestFileKey = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.MANIFEST);
 
         try (final ByteArrayInputStream manifestContent = new ByteArrayInputStream(manifest.getBytes())) {
-            uploader.upload(manifestContent, manifestFileKey);
+            final var bytes = uploader.upload(manifestContent, manifestFileKey);
+            metrics.recordObjectUpload(
+                remoteLogSegmentMetadata.remoteLogSegmentId().topicIdPartition().topicPartition(),
+                ObjectKey.Suffix.MANIFEST,
+                bytes
+            );
         }
     }
 
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
@@ -30,9 +30,13 @@ import org.apache.kafka.common.metrics.stats.Max;
 import org.apache.kafka.common.metrics.stats.Rate;
 import org.apache.kafka.common.utils.Time;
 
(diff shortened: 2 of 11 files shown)


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -46,12 +46,14 @@ public void configure(final Map<String, ?> configs) {
     }
 
     @Override
-    public void upload(final InputStream inputStream, final String key) throws StorageBackendException {
-        try (final var out = s3OutputStream(key)) {
+    public long upload(final InputStream inputStream, final String key) throws StorageBackendException {
+        final S3MultiPartOutputStream out = s3OutputStream(key);
+        try (out) {
```
ivanyu, line 51: The constructor may throw `AmazonServiceException`, so it still needs to be inside the `try`.

ivanyu (changes requested):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -46,9 +46,23 @@ public void configure(final Map<String, ?> configs) {
...
+        }
+    }
+
+    void transferContent(final String key,
+                         final InputStream in,
+                         final S3MultiPartOutputStream out)
+        throws StorageBackendException {
+        try {
+            in.transferTo(out);
+            out.close();
```
ivanyu, line 65: We can add one line to `upload` to make it return the size:
```java
public long upload(final InputStream inputStream, final String key) throws StorageBackendException {
    try (final var out = s3OutputStream(key)) {
        inputStream.transferTo(out);
        return out.uploadedBytes();
    } catch (final AmazonS3Exception | IOException e) {
        throw new StorageBackendException("Failed to upload " + key, e);
    }
}
```

Anything I miss that make it necessary to split it like this and abandon try-with-resources?

jeqo: In the meantime: I was counting bytes on flush which depended on close (last flush happens there). But I have moved the increment a bit earlier to the write method, so we can get the size as it's transferred to the buffer but not flushed necessarily.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -204,7 +203,7 @@ public void copyLogSegmentData(final RemoteLogSegmentMetadata remoteLogSegmentMe
                 final InputStream transactionIndex = Files.newInputStream(logSegmentData.transactionIndex().get());
                 uploadIndexFile(remoteLogSegmentMetadata, transactionIndex, TRANSACTION);
             }
-            final ByteBufferInputStream leaderEpoch = new ByteBufferInputStream(logSegmentData.leaderEpochIndex());
+            final InputStream leaderEpoch = new ByteArrayInputStream(logSegmentData.leaderEpochIndex().array());
```
giuseppelillo, line 206: Is there any particular reason for this change?

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
```diff
@@ -102,6 +106,26 @@ public Metrics(final Time time) {
         segmentFetchRequestedBytes.add(
             metrics.metricName("segment-fetch-requested-bytes-total", metricGroup),
             new CumulativeSum());
+
+        final var objectUploadRequestsRate =
+            new MetricNameTemplate("object-upload-rate", metricGroup, "", "object-type");
```
giuseppelillo, line 111: Could we add a description here?

jeqo: Good catch. Others metrics have no description neither, we could tackle this on a separate PR I think.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: Good question. Metric tests were failing with ByteBuffer input stream, and though there may be some issue with the state of the byte buffer state received at this point.
Looking deeper though, it was the source byteBuffer causing this issue. So by flipping it after each copy in the Metrics test it's passing correctly. 

So, rolling back this change, to avoid copying byte arrays. Also checked that resulting leader epoch is correctly uploaded. Pushing changes now.

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
jeqo: https://github.com/aiven/tiered-storage-for-apache-kafka/issues/313

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -238,17 +238,20 @@ private void uploadSegmentLog(final RemoteLogSegmentMetadata remoteLogSegmentMet
         throws IOException, StorageBackendException {
         final String fileKey = objectKey.key(remoteLogSegmentMetadata, ObjectKey.Suffix.LOG);
         try (final var sis = transformFinisher.toInputStream()) {
-            uploader.upload(sis, fileKey);
+            final var bytes = uploader.upload(sis, fileKey);
+            metrics.recordObjectUpload(ObjectKey.Suffix.LOG, bytes);
```
ivanyu, line 242: We need a way to get the aggregated number of bytes. It seems to me, one of:
1. Add a separate aggregated sensor.
2. Make the suffix a tag instead of a part of the sensor name (so we can sum by this tag).
3. Don't bother with separating by tag, the log files will anyway be 99%.

jeqo: We can follow a similar approach as with topic vs topic-partition, and have an aggregated metric on INFO level, and by object type in DEBUG, wdyt?

jeqo: Sure, giving a try now. Pushed latest changes.

ivanyu (approved):

PR #282, opened 2023-06-14
feat(storage:s3): abort on stream exception handling

jeqo:
Ensure multi-part uploads are aborted when exception happens anywhere on the stream processing.


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -47,9 +47,12 @@ public void configure(final Map<String, ?> configs) {
 
     @Override
     public void upload(final InputStream inputStream, final String key) throws StorageBackendException {
-        try (final var out = s3OutputStream(key)) {
+        final S3MultiPartOutputStream out = s3OutputStream(key);
+        try {
             inputStream.transferTo(out);
+            out.close();
```
AnatolyPopov, line 53: Why not try-finally?

jeqo: Want to avoid throwing IllegalStateException as stream is already closed after calling abort.

AnatolyPopov: So abort in the catch block guarantees that it will be closed? That should be documented at least I guess

jeqo: Agree, adding some notes to the class.

AnatolyPopov: Is it covered with tests also?

jeqo: Good catch. Adding a test case on S3StorageTest

ivanyu: > Want to avoid throwing IllegalStateException as stream is already closed after calling abort.

If throwing `IllegalStateException` breaks basic language idioms, probably it's not a good idea to do this. In most (all?) standard `InputStream`s `close` is idempotent and can be called multiple times. We probably should follow

jeqo: @ivanyu good point. I think close validation can be relaxed and only log and avoid side effects instead of throwing exceptions.

AnatolyPopov: Was thinking about exactly what @ivanyu said in the last comment.

jeqo:
OutputStream refactored to handle aborts within the processing and keep the close more idiomatic. Let me know how it looks now. 
cc @ivanyu @AnatolyPopov 
PS. I just saw your comment now, hopefully there's no need for an additional flag with the latest approach, but let me know. Thanks!

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -56,7 +55,7 @@ public class S3MultiPartOutputStream extends OutputStream {
     private final String uploadId;
     private final List<PartETag> partETags = new ArrayList<>();
 
-    private boolean closed;
+    private volatile boolean closed;
```
ivanyu, line 58: Should we do this? This class is not thread-safe and is supposed to be used inside one thread anyway.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +85,72 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
     }
 
     @Override
     public void close() throws IOException {
-        if (partBuffer.position() > 0) {
-            flushBuffer(partBuffer.arrayOffset(), partBuffer.position());
-        }
-        if (Objects.nonNull(uploadId)) {
+        if (closed) {
+            log.warn("Already closed");
```
ivanyu, line 110: I think we shouldn't even have this logging: this is not a condition we should be warned about and we can't do nothing about it. IMO just `return` is totally sufficient here.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +85,72 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
+    }
+
+    private CompleteMultipartUploadResult completeUpload() {
+        final var request = new CompleteMultipartUploadRequest(bucketName, key, uploadId, partETags);
+        final var result = client.completeMultipartUpload(request);
+        closed = true;
+        return result;
+    }
+
+    void abortUpload() {
```
ivanyu, line 143:
```suggestion
    private void abortUpload() {
```

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
ivanyu, line 1: I'd propose to add this test:
```java
@Test
void sendAbortIfNoWritingHappened() throws IOException {
    doNothing().when(mockedS3).abortMultipartUpload(abortMultipartUploadRequestCaptor.capture());

    try (final S3MultiPartOutputStream out =
             new S3MultiPartOutputStream(BUCKET_NAME, FILE_KEY, 100, mockedS3)) {
    }
    verify(mockedS3).abortMultipartUpload(abortMultipartUploadRequestCaptor.getValue());
    assertAbortMultipartUploadRequest(abortMultipartUploadRequestCaptor.getValue());
}
```

AnatolyPopov: The mocking is weird, capture should be in verify. otherwise you are expecting a call to be made with a captor as argument AFAIU.

AnatolyPopov: Also the verify checks that the method was called with captured value? This is super weird.

ivanyu: https://github.com/aiven/tiered-storage-for-apache-kafka/pull/285

jeqo: Thanks @ivanyu -- I think we don't need the `doNothing()...`, and agree with @AnatolyPopov that verify should use capture instead of getValue. Adding this on the next commit.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +85,72 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
+                    throw new IOException(e);
+                }
+            }
             if (!partETags.isEmpty()) {
                 try {
-                    final CompleteMultipartUploadRequest request =
-                        new CompleteMultipartUploadRequest(bucketName, key, uploadId, partETags);
-                    final CompleteMultipartUploadResult result = client.completeMultipartUpload(request);
+                    final CompleteMultipartUploadResult result = completeUpload();
                     log.debug("Completed multipart upload {} with result {}", uploadId, result);
```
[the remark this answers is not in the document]
jeqo: Good catch, thanks @ivanyu -- applying this suggestion to the next commit.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jeqo: Got it. Was getting worried as this value changes in multiple places.
Will add a note about this to the stream, and I'm extending tests to validate this value.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -231,9 +231,41 @@ void writesTailMessages() throws Exception {
...
+        verify(mockedS3, times(1)).completeMultipartUpload(completeMultipartUploadRequestCaptor.capture());
         assertCompleteMultipartUploadRequest(completeMultipartUploadRequestCaptor.getValue(),
             List.of(new PartETag(1, "SOME_ETAG#1"), new PartETag(2, "SOME_ETAG#2")));
+
+        assertStreamIsClosed(out);
+    }
+
+    @Test
+    void sendAbortIfNoWritingHappened() throws IOException {
+        when(mockedS3.initiateMultipartUpload(any(InitiateMultipartUploadRequest.class)))
```
AnatolyPopov, line 243: Nit: any()?

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -231,9 +231,41 @@ void writesTailMessages() throws Exception {
...
+        out.close();
+
+        verify(mockedS3).abortMultipartUpload(abortMultipartUploadRequestCaptor.capture());
+        assertAbortMultipartUploadRequest(abortMultipartUploadRequestCaptor.getValue());
+        assertStreamIsClosed(out);
+    }
+
+    @Test
+    void failWhenUploadingPartAfterStreamIsClosed() throws IOException {
+        when(mockedS3.initiateMultipartUpload(any(InitiateMultipartUploadRequest.class)))
```
AnatolyPopov, line 256: Nit: any()?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +86,73 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
+                partBuffer.put(source.array(), offset, transferred);
+                source.position(source.position() + transferred);
+                if (!partBuffer.hasRemaining()) {
+                    flushBuffer(0, partSize);
+                }
             }
+        } catch (final Exception e) {
+            log.error("Failed to write to stream on upload {}, aborting transaction", uploadId, e);
+            abortUpload();
+            throw new IOException(e);
```
AnatolyPopov, line 104: Why IOException? And don't see this covered by tests

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +86,73 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
-        }
-        if (Objects.nonNull(uploadId)) {
+        if (!closed) {
+            if (partBuffer.position() > 0) {
+                try {
+                    flushBuffer(partBuffer.arrayOffset(), partBuffer.position());
+                } catch (final Exception e) {
+                    log.error("Failed to upload last part {}, aborting transaction", uploadId, e);
+                    abortUpload();
+                    throw new IOException(e);
```
AnatolyPopov, line 117: Same: why IOException? Test coverage?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +86,73 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
                 }
             } else {
-                client.abortMultipartUpload(new AbortMultipartUploadRequest(bucketName, key, uploadId));
+                abortUpload();
             }
         }
+    }
+
+    public boolean isClosed() {
+        return closed;
```
AnatolyPopov, line 136: If I replace this with just true the tests are passing. Should we use this method inside the class also instead of field directly?

AnatolyPopov (changes requested):
Some comments

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +86,73 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
     public void close() throws IOException {
-        if (partBuffer.position() > 0) {
-            flushBuffer(partBuffer.arrayOffset(), partBuffer.position());
-        }
-        if (Objects.nonNull(uploadId)) {
+        if (!closed) {
+            if (partBuffer.position() > 0) {
+                try {
+                    flushBuffer(partBuffer.arrayOffset(), partBuffer.position());
+                } catch (final Exception e) {
```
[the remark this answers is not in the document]
jeqo: Sure, adding this.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -86,54 +86,73 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
+    }
+
+    private void abortUpload() {
+        final var request = new AbortMultipartUploadRequest(bucketName, key, uploadId);
+        client.abortMultipartUpload(request);
         closed = true;
     }
 
     private void flushBuffer(final int offset,
                              final int actualPartSize) throws IOException {
```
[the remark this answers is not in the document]
jeqo: Yes, good catch. Removing this.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -282,4 +314,9 @@ private static void assertAbortMultipartUploadRequest(final AbortMultipartUpload
         assertThat(request.getKey()).isEqualTo(FILE_KEY);
         assertThat(request.getUploadId()).isEqualTo(UPLOAD_ID);
     }
+
+    private void assertStreamIsClosed(final S3MultiPartOutputStream out) throws IOException {
+        assertThat(out.isClosed()).isTrue();
+        out.close();
```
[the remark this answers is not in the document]
jeqo: Added some comment and assertion to this call. The goal is to verify no additional calls to complete/abort/upload are done if close is recalled.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -77,11 +77,10 @@ void sendAbortForAnyExceptionWhileWriting() {
             .thenThrow(testException);
 
         assertThatThrownBy(() -> {
-            try (final S3MultiPartOutputStream out =
-                     new S3MultiPartOutputStream(BUCKET_NAME, FILE_KEY, 100, mockedS3)) {
+            try (final var out = new S3MultiPartOutputStream(BUCKET_NAME, FILE_KEY, 100, mockedS3)) {
                 out.write(new byte[] {1, 2, 3});
             }
-        }).isInstanceOf(IOException.class).hasCause(testException);
+        }).isInstanceOf(IOException.class).hasRootCause(testException);
```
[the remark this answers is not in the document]
jeqo: I think it test the whole exception `testException` instance including the message; or I may be missing something?

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -124,13 +124,13 @@ void writesOneByte() throws Exception {
         when(mockedS3.completeMultipartUpload(any()))
             .thenReturn(new CompleteMultipartUploadResult());
 
-        try (final S3MultiPartOutputStream out = new S3MultiPartOutputStream(BUCKET_NAME, FILE_KEY, 100, mockedS3)) {
-            out.write(1);
-        }
+        final var out = new S3MultiPartOutputStream(BUCKET_NAME, FILE_KEY, 100, mockedS3);
+        out.write(1);
+        out.close();
```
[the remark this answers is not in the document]
jeqo: For the sake of having the new assertions for retry closing streams

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -159,18 +160,17 @@ void writesMultipleMessages() throws Exception {
...
+            random.nextBytes(message);
+            out.write(message, 0, message.length);
+            expectedMessagesList.add(message);
         }
+        out.close();
 
         verify(mockedS3).initiateMultipartUpload(any(InitiateMultipartUploadRequest.class));
         verify(mockedS3, times(3)).uploadPart(uploadPartRequestCaptor.capture());
-        verify(mockedS3).completeMultipartUpload(completeMultipartUploadRequestCaptor.capture());
+        verify(mockedS3, times(1)).completeMultipartUpload(completeMultipartUploadRequestCaptor.capture());
```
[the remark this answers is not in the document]
jeqo: Got it, removing it then.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
jeqo: Got caught up in the rebase, good catch, fixing it.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jeqo: I'm piggybacking on the existing checked exception. Also, first two tests are checking this type already when failing on writing or closing stream.
Let me know if I'm missing something.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jeqo: Good catch. Using this method on conditions

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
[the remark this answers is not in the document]
AnatolyPopov: Yep, my bad. :+1:

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -282,4 +320,10 @@ private static void assertAbortMultipartUploadRequest(final AbortMultipartUpload
         assertThat(request.getKey()).isEqualTo(FILE_KEY);
         assertThat(request.getUploadId()).isEqualTo(UPLOAD_ID);
     }
+
+    private void assertStreamIsClosed(final S3MultiPartOutputStream out) {
+        assertThat(out.isClosed()).isTrue();
+        // retry close to validate no exception is thrown and number of calls to complete/upload does not change
```
AnatolyPopov, line 326: Where do we actually verify the number of calls?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
AnatolyPopov: Still test coverage.

AnatolyPopov (commented):
A couple more nits

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
jeqo: it's on each test case as it depends on the test condition. 
i guess would it be better to pull those in here? can give a try.

jeqo: actually, taking this out to the methods itself. A bit of repeated code, but clearer to assess within each test case, wdyt?

AnatolyPopov: I agree but the comment is still misleading. After you added an assertions it says on it's own that we are testing that `close` does not throw exception. For the number of calls we are sometimes(not everywhere) are checking the expected number of calls for the whole test, not that the number of calls does not change(e.g. verifyNoMoreInterations or similar). So I would drop those comments altogether.

AnatolyPopov (approved):
LGTM, thanks for addressing all the comments!.

ivanyu (approved):

PR #283, opened 2023-06-14
Removing unnecessary additional storage initialization

AnatolyPopov:


core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -78,9 +76,6 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
 
     private final Executor executor = new ForkJoinPool();
 
-    private ObjectFetcher fetcher;
-    private ObjectUploader uploader;
-    private ObjectDeleter deleter;
```
ivanyu, line 83: From our discussions, we explicitly wanted these to be separate, primarily to have separate S3 clients for different purposes.

ivanyu (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
AnatolyPopov: Should then this be documented? I vaguely remember that but it's not obvious

AnatolyPopov: Adjusting the PR accordingly

AnatolyPopov: However I still think that we then need to create separate clients inside s3 implementation instead.

PR #285, opened 2023-06-15
Use captors correctly in tests

ivanyu:


storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -210,7 +210,7 @@ void writesTailMessages() throws Exception {
 
                 return newUploadPartResult(up.getPartNumber(), "SOME_TAG#" + up.getPartNumber());
             });
-        when(mockedS3.completeMultipartUpload(completeMultipartUploadRequestCaptor.capture()))
```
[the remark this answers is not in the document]
ivanyu: I added checking

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -153,12 +150,12 @@ void writesMultipleMessages() throws Exception {
 
         when(mockedS3.initiateMultipartUpload(any(InitiateMultipartUploadRequest.class)))
             .thenReturn(newInitiateMultipartUploadResult());
-        when(mockedS3.uploadPart(uploadPartRequestCaptor.capture()))
+        when(mockedS3.uploadPart(any()))
```
jeqo, line 153: For consistency: we were using `any(Type class)` before, should we stick to that?

ivanyu: Thanks, yes. Fixed

jeqo (approved):

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
AnatolyPopov: It depends on what we want to mock I guess. If we want to mock only the calls when the argument of specific type then @jeqo you are right. But in this case I think we need `all` calls including null.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -122,18 +119,18 @@ void sendAbortForAnyExceptionWhenClose() throws Exception {
     void writesOneByte() throws Exception {
         when(mockedS3.initiateMultipartUpload(any(InitiateMultipartUploadRequest.class)))
             .thenReturn(newInitiateMultipartUploadResult());
-        when(mockedS3.uploadPart(uploadPartRequestCaptor.capture()))
+        when(mockedS3.uploadPart(any(UploadPartRequest.class)))
```
AnatolyPopov, line 122: I guess class argument is not needed, see another thread. Same in other places.

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
ivanyu: Well, it'll work either way, because we don't expect `null` in these cases and the type will match exactly

AnatolyPopov: But if for some reason null will be passed then the test will fail with unused stubbing error instead of proper error message. If we do not expect null then we should check that the captured value is not null.

jeqo: This is mocking a response. It shouldn't return something when request is null, isn't?
Also, not sure where the null validation would go. The upload requests are validated already.

ivanyu: It will be `Strict stubbing argument mismatch`, not `Unused stubbing`:
```
Strict stubbing argument mismatch. Please check:
 - this invocation of 'uploadPart' method:
    mockedS3.uploadPart(null);
    -> at io.aiven.kafka.tieredstorage.storage.s3.S3MultiPartOutputStream.uploadPart(S3MultiPartOutputStream.java:149)
 - has following stubbing(s) with different arguments:
    1. mockedS3.uploadPart(null);
      -> at io.aiven.kafka.tieredstorage.storage.s3.S3MultiPartOutputStreamTest.sendAbortForAnyExceptionWhileWriting(S3MultiPartOutputStreamTest.java:76)
```
which is pretty clear.

AnatolyPopov: @jeqo The null validation can be just assert that the captured argument is not null. IMO we should not make assumptions on nullability int the test. At the same time since this mock is reused in different tests somebody at some point can decide that strictness of it should be lenient and the same time having null passed in production code for some reason. I know too many conditions. But I still believe that the test method should be self sufficient and not relying on the assumptions unrelated to it. And yes, I've seen this kind of failures simply introduced by wrongly rebased code.

jeqo: agreed offline that we don't need typed any(..) as there's no polymorphic methods tested, and decided to apply `any()` unless different types are expected.

jeqo (approved):

PR #290, opened 2023-06-16
Exclude Jackson libraries that are supposed to be provided by Kafka

ivanyu:


AnatolyPopov (approved):
LGTM

ivanyu:
`ChildFirstClassLoader` should isolate well, so technically these exclusions are not needed. Parking this for a while

PR #291, opened 2023-06-19
feat(metrics): add metrics for delete path

jeqo:
Record metrics for:
- Delete requests (req/sec)
- Bytes requested to delete (bytes/sec)

This is based on RSM metadata, and not actual deletes. Those would be measured at the storage layer.


ivanyu:
It's worth adding the total metrics, too. If you prefer, we can merge this first

PR #292, opened 2023-06-19
feat(metrics): additional metrics

jeqo:
Changes:

- Add segment delete time metric
- Add total metric to segment fetch requests
- Add fetch bytes metric

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -270,15 +270,15 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
     public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
                                        final int startPosition,
                                        final int endPosition) throws RemoteStorageException {
-        metrics.recordSegmentFetch();
-
         try {
-            final SegmentManifest segmentManifest = segmentManifestProvider.get(remoteLogSegmentMetadata);
-
             final BytesRange range = BytesRange.of(
                 startPosition,
                 Math.min(endPosition, remoteLogSegmentMetadata.segmentSizeInBytes() - 1)
             );
+            metrics.recordSegmentFetch(range.size());
+
+            final SegmentManifest segmentManifest = segmentManifestProvider.get(remoteLogSegmentMetadata);
+
             return new FetchChunkEnumeration(chunkManager, remoteLogSegmentMetadata, segmentManifest, range)
                 .toInputStream();
         } catch (final StorageBackendException | IOException e) {
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -308,6 +308,8 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
         throws RemoteStorageException {
         metrics.recordSegmentDelete(remoteLogSegmentMetadata.segmentSizeInBytes());
 
+        final long startedMs = time.milliseconds();
+
         try {
             for (final ObjectKey.Suffix suffix : ObjectKey.Suffix.values()) {
                 final String key = objectKey.key(remoteLogSegmentMetadata, suffix);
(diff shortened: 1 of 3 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -202,9 +202,9 @@ public void copyLogSegmentData(final RemoteLogSegmentMetadata remoteLogSegmentMe
             uploadIndexFile(remoteLogSegmentMetadata, leaderEpoch, LEADER_EPOCH);
         } catch (final StorageBackendException | IOException e) {
             throw new RemoteStorageException(e);
+        } finally {
+            metrics.recordSegmentCopyTime(startedMs, time.milliseconds());
         }
-
-        metrics.recordSegmentCopyTime(startedMs, time.milliseconds());
```
ivanyu, line 207: We need to decide if we count failed attempts here and in other places. In the current code they aren't counted as the `record*` is called outside of `try`.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -270,19 +270,23 @@ public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegme
...
+        final long startedMs = time.milliseconds();
 
         try {
-            final SegmentManifest segmentManifest = segmentManifestProvider.get(remoteLogSegmentMetadata);
-
             final BytesRange range = BytesRange.of(
                 startPosition,
                 Math.min(endPosition, remoteLogSegmentMetadata.segmentSizeInBytes() - 1)
             );
+            metrics.recordSegmentFetch(range.size());
```
ivanyu, line 280: This will be skewed almost always: we request from `startPosition` to the end of file, but read only so many bytes on the broker side. We probably need this metric inside some wrapper `InputStream` that counts read bytes and register on close.

jeqo: Agree, though I think we need both. We can have this as the broker-side request (hopefully with proper ranges in the short future), and at the input stream level as you suggest to study both metrics, wdyt?

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: Sure, we can take this on a separate PR

jeqo:
Removing `fetch-latency` as it doesn't give much meaning at the moment, as there is not measuring of how input stream is being read.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
AnatolyPopov: If we want to keep this metric here I would suggest to rename it to something like `segment-fetch-requested` or `segment-fetch-bytes-requested` or whatever else saying that this is the amount of bytes that was requested by broker but neither the amount of bytes that was in the end fetched from remote storage nor the amount of bytes that was actually read by the broker.

core/src/test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerMetricsTest.java
```diff
@@ -105,38 +105,50 @@ void metricsShouldBeReported() throws RemoteStorageException, JMException {
...
-        assertThat((double) MBEAN_SERVER.getAttribute(segmentCopyPerSecName, "segment-copy-time-avg"))
-            .isZero();
-        assertThat((double) MBEAN_SERVER.getAttribute(segmentCopyPerSecName, "segment-copy-time-max"))
-            .isZero();
-
-        assertThat((double) MBEAN_SERVER.getAttribute(segmentCopyPerSecName, "segment-copy-bytes-rate"))
+        assertThat((double) MBEAN_SERVER.getAttribute(rsmMetricsName, "segment-copy-bytes-rate"))
             .isEqualTo(30.0 / METRIC_TIME_WINDOW_SEC);
-        assertThat((double) MBEAN_SERVER.getAttribute(segmentCopyPerSecName, "segment-copy-bytes-total"))
+        assertThat((double) MBEAN_SERVER.getAttribute(rsmMetricsName, "segment-copy-bytes-total"))
```
AnatolyPopov, line 117: Do we really need to cast to double everywhere? `isZero` will not work without cast but `isEqualTo(0.0)` will in case you want to keep common style throughout the method.

AnatolyPopov (commented):
Some more comments

ivanyu (approved):

PR #293, opened 2023-06-20
feat(docker): add jmx exporter to expose prometheus metrics

jeqo:


docker/Dockerfile
```diff
@@ -24,5 +24,13 @@ RUN cd /tiered-storage-for-apache-kafka \
     && tar -xf tiered-storage-for-apache-kafka-${_VERSION}.tgz --strip-components=1 \
     && rm tiered-storage-for-apache-kafka-${_VERSION}.tgz
 
+# Installing JMX exporter agent
+ENV JMX_EXPORTER_VERSION=0.18.0
```
ivanyu, line 28: This probably should be ARG instead. Otherwise, it's possible to build with one version and then to run with another (without success).

demo/Makefile
```diff
@@ -27,7 +27,7 @@ local_retention_ms=1000 # 1 second
 
 .PHONY: create_topic
 create_topic:
-	docker exec $(kafka_container) \
+	docker exec -e KAFKA_OPTS= $(kafka_container) \
```
ivanyu, line 30: I suspect `KAFKA_OPTS` is taken into account only when a real Kafka is started via the `/etc/confluent/docker/configure` script, so this probably can be skipped here and below

ivanyu (changes requested):

demo/Makefile
jeqo: Nope, it's picked by the `kafka-run.sh` command, some it tries to initialize the agent on every execution (unfortunately).

ivanyu: Yeah, you're right. It's Kafka's own thing, not Conlfuent's.

In this case, could you please add the same to `create_topic_ts_by_time`, `create_topic_no_ts_by_size`, `create_topic_no_ts_by_time`?

jeqo: Of course 😅 good catch

ivanyu (approved):

Issue #287, opened 2023-06-15
[New implementation] The fetchLogSegment implementation is flawed


ivanyu:
The broker side will close the `InputStream` returned from `fetchLogSegment`. This is currently `SequenceInputStream`, which on `close()` will iterate over the sub-`InputStream`s. Each of these sub-streams will initiate a GET to the remote storage. This causes long delays, excess traffic and pollutes the cache.

We need to drop `SequenceInputStream` and replace it with our own `ChunkManager`-based `InputStream` implementation.

PR #294, opened 2023-06-20
Getting rid of unnecessary chunk fetching

AnatolyPopov:
Resolves: #287

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/transform/FetchChunkEnumeration.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/transform/FetchChunkEnumeration.java
@@ -44,6 +44,7 @@ public class FetchChunkEnumeration implements Enumeration<InputStream> {
     final int lastChunkId;
     private final ChunkIndex chunkIndex;
     int currentChunkId;
+    public boolean closed;
 
     /**
      *
@@ -92,7 +93,7 @@ public class FetchChunkEnumeration implements Enumeration<InputStream> {
 
     @Override
     public boolean hasMoreElements() {
-        return currentChunkId <= lastChunkId;
+        return !closed && currentChunkId <= lastChunkId;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/transform/CloseableEnumeration.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage.transform;
+
+public abstract class CloseableEnumeration<T> implements java.util.Enumeration<T>, AutoCloseable {
```
jeqo, line 19:
```suggestion
public abstract class CloseableEnumeration<T extends InputStream> implements java.util.Enumeration<T>, AutoCloseable {
```

core/src/main/java/io/aiven/kafka/tieredstorage/transform/FetchChunkEnumeration.java
```diff
@@ -35,7 +33,7 @@
 
 import org.apache.commons.io.input.BoundedInputStream;
 
-public class FetchChunkEnumeration implements Enumeration<InputStream> {
+public class FetchChunkEnumeration extends CloseableEnumeration<InputStream> {
```
jeqo, line 36: Aren't the other enumerations also affected by this issue? If they aren't, maybe we can document on the code why are these proposed wrappers required.

jeqo (commented):
A couple of suggestions.

core/src/main/java/io/aiven/kafka/tieredstorage/transform/CloseableEnumeration.java
AnatolyPopov: I don't think we should have tied to `InputStream` specifically since it can be reused for whatever else types in future if needed.

jeqo (approved):
LGTM

ivanyu (commented):
Great, thanks, this should do

PR #309, opened 2023-07-03
feat(metrics): tag RSM metrics by topic and partition

jeqo:
Changes:

- Add topic and partition label to existing metrics.
- Refactor the metrics module to provide metric sensor if already exists.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
@@ -21,12 +21,17 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 
+import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.metrics.Sensor;
 import org.apache.kafka.common.utils.Utils;
 
 import io.aiven.kafka.tieredstorage.storage.StorageBackend;
+
+import static org.apache.kafka.common.config.ConfigDef.Range.atLeast;
+import static org.apache.kafka.common.config.ConfigDef.ValidString.in;
 
 public class RemoteStorageManagerConfig extends AbstractConfig {
     private static final String STORAGE_PREFIX = "storage.";
@@ -68,6 +73,17 @@ public class RemoteStorageManagerConfig extends AbstractConfig {
     private static final String ENCRYPTION_PRIVATE_KEY_FILE_CONFIG = "encryption.private.key.file";
     private static final String ENCRYPTION_PRIVATE_KEY_FILE_DOC = "The path to the RSA private key file";
     // TODO add possibility to pass keys as strings
+
+
+    public static final String METRICS_NUM_SAMPLES_CONFIG = CommonClientConfigs.METRICS_NUM_SAMPLES_CONFIG;
+    private static final String METRICS_NUM_SAMPLES_DOC = CommonClientConfigs.METRICS_NUM_SAMPLES_DOC;
+
+    public static final String METRICS_SAMPLE_WINDOW_MS_CONFIG = CommonClientConfigs.METRICS_SAMPLE_WINDOW_MS_CONFIG;
+    private static final String METRICS_SAMPLE_WINDOW_MS_DOC = CommonClientConfigs.METRICS_SAMPLE_WINDOW_MS_DOC;
+
+    public static final String METRICS_RECORDING_LEVEL_CONFIG = CommonClientConfigs.METRICS_RECORDING_LEVEL_CONFIG;
+    private static final String METRICS_RECORDING_LEVEL_DOC = CommonClientConfigs.METRICS_RECORDING_LEVEL_DOC;
+
 
     private static final ConfigDef CONFIG;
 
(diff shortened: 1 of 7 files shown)


ivanyu:
It's probably worth keeping the aggregated metrics as well, e.g. like Kafka [does](https://github.com/apache/kafka/blob/928070e3206867fa81bafbb138373bcb0e2ba962/core/src/main/scala/kafka/server/ReplicaManager.scala#L1052-L1053), for convenience.

ivanyu:
Should we have the partition-level breakdown? Kafka mostly reports on the topic level.

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/SensorProvider.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private final Sensor sensor;
+
+    private final boolean preexisting;
+
+    private final Map<String, String> tags;
+
+    public SensorProvider(final Metrics metrics,
+                          final String name,
+                          final Supplier<Map<String, String>> tagsSupplier) {
```
ivanyu, line 42: Does it need to be a function?

jeqo: I think is a small optimization to avoid calling the supplier when sensor already exists.

ivanyu:
Yeah, the levels may help here. We can have some config like the `metrics.recording.level` from various parts of Kafka.

ivanyu:
If it fits nicely into one PR, probably just feel free to modify this one. Otherwise, it's OK to split (i.e. implement only the topic-level here + make a ticket for later). How you see better

jeqo:
Sure! Adding metric configs and metrics for all and tagged only by topic

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
```diff
@@ -156,6 +172,28 @@ public class RemoteStorageManagerConfig extends AbstractConfig {
             ConfigDef.Importance.HIGH,
             ENCRYPTION_PRIVATE_KEY_FILE_DOC
         );
+
+        CONFIG
+            .define(METRICS_SAMPLE_WINDOW_MS_CONFIG,
+                ConfigDef.Type.LONG,
+                30000,
+                atLeast(0),
```
ivanyu, line 180: I guess it doesn't make sense to be 0
```suggestion
                atLeast(1),
```

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/MetricsRegistry.java
```diff
@@ -0,0 +1,268 @@
...
+    static final String SEGMENT_FETCH_REQUESTED_BYTES_RATE = SEGMENT_FETCH_REQUESTED_BYTES + "-rate";
+    final MetricNameTemplate segmentFetchRequestedBytesRate;
+    final MetricNameTemplate segmentFetchRequestedBytesRateByTopic;
+    final MetricNameTemplate segmentFetchRequestedBytesRateByTopicPartition;
+    static final String SEGMENT_FETCH_REQUESTED_BYTES_TOTAL = SEGMENT_FETCH_REQUESTED_BYTES + "-total";
+    final MetricNameTemplate segmentFetchRequestedBytesTotal;
+    final MetricNameTemplate segmentFetchRequestedBytesTotalByTopic;
+    final MetricNameTemplate segmentFetchRequestedBytesTotalByTopicPartition;
+
+    public MetricsRegistry() {
```
ivanyu, line 125: Not sure if it's worth changing now, but all the fields seemingly can be initialized in place and the constructor is not needed

ivanyu:
LGTM apart from the comments (and the second one is up to you)

jeqo:
Thanks @ivanyu ! Suggestion sounds good to me, pushed latest version

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
```diff
@@ -32,107 +33,267 @@
...
+            .get()
+            .record(time);
+    }
 
-        segmentDeleteTime = metrics.sensor("segment-delete-time");
-        segmentDeleteTime.add(metrics.metricName("segment-delete-time-avg", metricGroup), new Avg());
-        segmentDeleteTime.add(metrics.metricName("segment-delete-time-max", metricGroup), new Max());
+    public void recordSegmentCopyError(final TopicPartition topicPartition) {
+        recordSegmentCopyErrors(topicPartition);
+    }
```
ivanyu, line 138: Do we need this function or `recordSegmentCopyErrors` can be just made public?

jeqo: Sure, removing inner method.

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/Metrics.java
```diff
@@ -32,107 +33,259 @@
...
+            .with(metricsRegistry.segmentDeleteTimeMaxByTopicPartition, new Max())
+            .get()
+            .record(time);
     }
 
-    public void recordSegmentDelete(final int bytes) {
-        segmentDeleteRequests.record();
-        segmentDeleteBytes.record(bytes);
+    public void recordSegmentDeleteError(final TopicPartition topicPartition) {
+        new SensorProvider(metrics, sensorName(SEGMENT_DELETE_ERRORS))
```
ivanyu, line 227: Should probably be `SEGMENT_DELETE_ERRORS_RATE`? And below too

ivanyu: If so, probably `SEGMENT_DELETE_ERRORS` and other "base" constants should be made private to prevent this misuse

ivanyu: And also since the tests missed this too, maybe something is wrong with them

jeqo: It's correct actually. `SEGMENT_DELETE_ERRORS` is the sensor name, and `SEGMENT_DELETE_ERRORS_RATE` the metric name. `SEGMENT_DELETE_ERRORS_RATE` may be private. I may also add `METRIC_` and `SENSOR_` to make these names less confusing

ivanyu: No, I'm wrong :)

ivanyu (approved):

PR #310, opened 2023-07-03
feat(transform): support no chunking to encrypt indexes

jeqo:
Changes:
- Extend transform/detransform to avoid chunking.
- Use transform/detransfom on RSM to store and serve indexes


ivanyu:
@jeqo LGTM in general. Just wanted to check one idea with you: instead of explicit chunk-less mode in transforms, we can use a single surrogate chunk of `Integer.MAX_VALUE` size. What do you think of this? But anyway, the current approach is good and I'm ready to merge if you also like it better.

jeqo:
Good point, that would achieve the same; however we may need to apply a config validation for chunk size to be at least 1.

I like this approach slightly better though as it's easier to configure for users as well (setting zero compared to max value).

ivanyu (approved):

PR #311, opened 2023-07-04
fix(cache:disk): fail when base directory does not exist

jeqo:
Currently, disk-based cache is only validating if directory exist to delete it and re-create child directories.

This approach is inconsistent with how other paths are defined in the plugin, requiring the directory to exist prior to the start of the plugin. The plugin does this to avoid messing with the existing permissions around the preparation of the directory.

Applying the fixes to fail if directory doesn't exist and some additional refactoring.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactoryConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkManagerFactoryConfig.java
@@ -49,6 +49,7 @@ public class ChunkManagerFactoryConfig extends AbstractConfig {
         super(CONFIG, originals);
     }
 
+    @SuppressWarnings("unchecked")
     public Class<ChunkCache<?>> cacheClass() {
         return (Class<ChunkCache<?>>) getClass(CHUNK_CACHE_CONFIG);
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -104,7 +104,16 @@ TBD
...
+Segment chunks are cached locally to improve the fetching performance; as subsequent fetch requests will most likely hit the same chunk until all records are consumed.
+
+This cache is currently implemented as:
+- In-memory
+- Disk-based
+
+#### Disk-based Chunk Cache
+
+Requires a base directory (`path` configuration). 
+This directory is not created by the plugin, so it needs to be provided before starting the broker.
```
ivanyu, line 116: I think we shouldn't put this paragraph in this section, because it's the design overview, not the usage documentation. We can either create the usage section with this paragraph and lots of TBD, or drop it for now until we start writing the documentation.

ivanyu (changes requested):
LGTM in general, one small thing with the README

README.md
jeqo: Sure, created https://github.com/aiven/tiered-storage-for-apache-kafka/issues/312 to follow up, and dropped changes on README.md

ivanyu (approved):

PR #320, opened 2023-07-14
Identify encryption key in manifest

ivanyu:
This commit is a preparation for changing how the plugin deals with encryption keys. The idea will be that there will be several RSA key encryption keys (KEK) provided in the configuration. When a data encryption key (DEK) is encrypted and written to the manifest, it comes with the ID of the KEK that encrypted it. When it's time to decrypt, the corresponding KEK is found by ID in the keyring and used for decryption.

This commit doesn't yet introduce multiple keys on the configuration level, but set's up all the necessary internal mechanics and uses a static KEK ID for now.

There will be a follow-up that actually introduces the keyring on the configuration level.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -17,13 +17,13 @@
 package io.aiven.kafka.tieredstorage;
 
 import javax.crypto.SecretKey;
-import javax.crypto.spec.SecretKeySpec;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
+import java.security.KeyPair;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.Executor;
@@ -50,6 +50,7 @@ import io.aiven.kafka.tieredstorage.metrics.Metrics;
 import io.aiven.kafka.tieredstorage.security.AesEncryptionProvider;
 import io.aiven.kafka.tieredstorage.security.DataKeyAndAAD;
 import io.aiven.kafka.tieredstorage.security.RsaEncryptionProvider;
+import io.aiven.kafka.tieredstorage.security.RsaKeyReader;
 import io.aiven.kafka.tieredstorage.storage.BytesRange;
 import io.aiven.kafka.tieredstorage.storage.ObjectDeleter;
 import io.aiven.kafka.tieredstorage.storage.ObjectFetcher;
@@ -102,6 +103,8 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
 
     private SegmentManifestProvider segmentManifestProvider;
 
+    private static final String KEY_ENCRYPTION_KEY_ID = "static-key-id";
+
     public RemoteStorageManager() {
         this(Time.SYSTEM);
     }
@@ -124,10 +127,11 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
         objectKey = new ObjectKey(config.keyPrefix());
         encryptionEnabled = config.encryptionEnabled();
         if (encryptionEnabled) {
-            rsaEncryptionProvider = RsaEncryptionProvider.of(
-                config.encryptionPublicKeyFile(),
-                config.encryptionPrivateKeyFile()
-            );
+            final KeyPair keyPair =
+                RsaKeyReader.read(config.encryptionPublicKeyFile(), config.encryptionPrivateKeyFile());
+            rsaEncryptionProvider = new RsaEncryptionProvider(
+                KEY_ENCRYPTION_KEY_ID,
+                Map.of(KEY_ENCRYPTION_KEY_ID, keyPair));
             aesEncryptionProvider = new AesEncryptionProvider();
         }
         final ChunkManagerFactory chunkManagerFactory = new ChunkManagerFactory();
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/security/RsaKeyReader.java
ivanyu, line 1: This is extracted from `RsaEncryptionProvider` with some minor changes.

core/src/main/java/io/aiven/kafka/tieredstorage/security/RsaEncryptionProvider.java
```diff
@@ -112,54 +100,4 @@ private Cipher createDecryptingCipher(final Key key) {
             throw new RuntimeException("Couldn't create decrypt cipher", e);
         }
     }
-
-    static class RsaKeysReader {
-
```
ivanyu, line 117: Moved to a separate class

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -164,7 +169,7 @@ private ObjectMapper getObjectMapper() {
             final SimpleModule simpleModule = new SimpleModule();
             simpleModule.addSerializer(SecretKey.class, new DataKeySerializer(rsaEncryptionProvider::encryptDataKey));
             simpleModule.addDeserializer(SecretKey.class, new DataKeyDeserializer(
-                b -> new SecretKeySpec(rsaEncryptionProvider.decryptDataKey(b), "AES")));
+                (keyId, b) -> new SecretKeySpec(rsaEncryptionProvider.decryptDataKey(keyId, b), "AES")));
```
jeqo, line 172: nit: wondering if we can move the `SecretKeySpec` definition into the `DataKeyDeserializer` 

```suggestion
            simpleModule.addDeserializer(SecretKey.class, new DataKeyDeserializer(rsaEncryptionProvider::decryptDataKey));
```

not directly related to the change, but we can take the chance to improve it.

`DataKeyDeserializer` would have a 

```
    private final BiFunction<String, byte[], byte[]> keyDecryptor;
```
instead, and return
```
        final var decryptedKey = keyDecryptor.apply(keyEncryptionKeyId, Base64.getDecoder().decode(dataKeyBase64));
        return new SecretKeySpec(decryptedKey, "AES");
```

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/serde/DataKeyDeserializer.java
```diff
@@ -19,22 +19,38 @@
...
+            throw new JsonParseException("Malformed encrypted data key string: " + keyText);
+        }
+        // Sanity check for more than one ':'
+        final int lastColonI = keyText.lastIndexOf(':');
+        if (lastColonI != colonI) {
+            throw new JsonParseException("Malformed encrypted data key string: " + keyText);
+        }
+
+        final String keyEncryptionKeyId = keyText.substring(0, colonI);
+        final String dataKeyBase64 = keyText.substring(colonI + 1);
```
jeqo, line 53: What about moving this into `DataEncryptionKeyResult` instead?
Then we could have something like: 
```suggestion
        final var result = DataEncryptionKeyResult.decode(p.getText());
```

And BiFunction could be back to `Function<DataEncryptionKeyResult, byte[]>`

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/serde/DataKeySerializer.java
```diff
@@ -19,24 +19,32 @@
...
     @Override
     public void serialize(final SecretKey value,
                           final JsonGenerator gen,
                           final SerializerProvider provider) throws IOException {
-        gen.writeBinary(keyEncryptor.apply(value));
+        final DataKeyEncryptionResult encryptionResult = dataKeyEncryptor.apply(value);
+        gen.writeString(
+            encryptionResult.keyEncryptionKeyId
+            + ":"
+            + Base64.getEncoder().encodeToString(encryptionResult.encryptedDataKey));
```
jeqo, line 48: Following suggestion to move decoding to DataEncryptionKeyResult:
```suggestion
        gen.writeString(encryptionResult.encode());
```

jeqo (commented):
Looks good! Just a couple of suggestions

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
ivanyu: Yes, good idea. Done this, but also made `EncryptedDataKey` (ex-`DataEncryptionKeyResult`) passed between (de-)serializer and `RsaEncryptionProvider` both directions.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/serde/DataKeyDeserializer.java
ivanyu: Yes, done this. Also renamed it to `EncryptedDataKey`

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/serde/DataKeySerializer.java
ivanyu: Yes, done this. Also added more constructor checks and tests.

ivanyu:
@jeqo I done your proposed improvements + a couple more. Please let me know it's OK, I'll do a squash.

jeqo (approved):
LGTM, thanks @ivanyu

ivanyu:
Thanks @jeqo. Squashed

PR #323, opened 2023-07-18
feat(metrics): add cache metrics to chunk cache

jeqo:
Refactors Disk-based cache to report metrics, and includes Caffeine instrumentation to chunk-cache.

Depends on #322

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
@@ -34,6 +34,7 @@ import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
 import io.aiven.kafka.tieredstorage.chunkmanager.ChunkKey;
 import io.aiven.kafka.tieredstorage.chunkmanager.ChunkManager;
 import io.aiven.kafka.tieredstorage.manifest.SegmentManifest;
+import io.aiven.kafka.tieredstorage.metrics.CaffeineStatsCounter;
 import io.aiven.kafka.tieredstorage.storage.StorageBackendException;
 
 import com.github.benmanes.caffeine.cache.AsyncCache;
@@ -44,12 +45,18 @@ import com.github.benmanes.caffeine.cache.Weigher;
 
 public abstract class ChunkCache<T> implements ChunkManager, Configurable {
     private static final long GET_TIMEOUT_SEC = 10;
+    private static final String METRIC_GROUP = "chunk-cache";
+
     private final ChunkManager chunkManager;
     private final Executor executor = new ForkJoinPool();
+
+    final CaffeineStatsCounter statsCounter;
+
     protected AsyncCache<ChunkKey, T> cache;
 
     protected ChunkCache(final ChunkManager chunkManager) {
         this.chunkManager = chunkManager;
+        this.statsCounter = new CaffeineStatsCounter(METRIC_GROUP);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
```diff
@@ -60,38 +61,35 @@ protected ChunkCache(final ChunkManager chunkManager) {
...
-                    .thenApplyAsync(t -> result.get())
+                }, executor))
+                .get(GET_TIMEOUT_SEC, TimeUnit.SECONDS);
+            if (result.get() == null) {
+                return cacheMap.get(chunkKey)
+                    .thenApplyAsync(this::cachedChunkToInputStream)
                     .get(GET_TIMEOUT_SEC, TimeUnit.SECONDS);
+            } else {
+                return result.get();
+            }
```
AnatolyPopov, line 90: Can we avoid touching this altogether? It was specifically done like this to guarantee thread-safety + providing not a cached input stream but creating a new one on every get to the cache at the same time preserving the value of the cache to be InputStream. That brings some additional benefits under the hood also.

AnatolyPopov (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
jeqo: I wouldn't mind. Unfortunately, with `compute` there's no metric computation as the action is ad-hoc, and cannot count as hit or missed [1].
If we feel strongly about it, then will have to define custom metrics to be handled on top of this implementation 

[1] https://github.com/ben-manes/caffeine/issues/689

jeqo: After offline discussion, we are trying a new approach by calling StatCounter manually as part of the `compute` method.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
```diff
@@ -46,10 +47,12 @@ public abstract class ChunkCache<T> implements ChunkManager, Configurable {
     private static final long GET_TIMEOUT_SEC = 10;
     private final ChunkManager chunkManager;
     private final Executor executor = new ForkJoinPool();
+    final CaffeineStatsCounter statsCounter;
     protected AsyncCache<ChunkKey, T> cache;
 
     protected ChunkCache(final ChunkManager chunkManager) {
         this.chunkManager = chunkManager;
+        this.statsCounter = new CaffeineStatsCounter("chunk-cache");
```
AnatolyPopov, line 55: Should the name be a constant like in another place?

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheMetricsTest.java
```diff
@@ -0,0 +1,131 @@
...
+            .thenReturn(new ByteArrayInputStream("test".getBytes()));
+
+        diskBasedChunkCache.getChunk(REMOTE_LOG_SEGMENT_METADATA, segmentManifest, 0);
+        diskBasedChunkCache.getChunk(REMOTE_LOG_SEGMENT_METADATA, segmentManifest, 0);
+
+        final var objectName = new ObjectName("aiven.kafka.server.tieredstorage.cache:type=chunk-cache");
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-hits-total"))
+            .isEqualTo(1.0);
+        assertThat((double) MBEAN_SERVER.getAttribute(objectName, "cache-hits-rate"))
+            .isCloseTo(1.0 / METRIC_TIME_WINDOW_SEC, Percentage.withPercentage(99));
```
AnatolyPopov, line 101: Instead of all these class casts to `double` that could lead to potential NPE in case of null, should we use this?
```suggestion
        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-hits-rate"))
            .asInstanceOf(InstanceOfAssertFactories.DOUBLE)
            .isCloseTo(1.0 / METRIC_TIME_WINDOW_SEC, Percentage.withPercentage(99));
```
Might be a little bit too verbose however.

jeqo: Agree, it may be too verbose. We may have used double casting elsewhere. Would prefer to keep it if that's the case

AnatolyPopov:
I believe this one is partially dependent on https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/349

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheMetricsTest.java
```diff
@@ -0,0 +1,116 @@
...
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-load-success-time-total"))
+            .asInstanceOf(InstanceOfAssertFactories.DOUBLE)
+            .isGreaterThan(0);
+
+        // compute is considered as load success regardless if present or not
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-load-success-total"))
+            .isEqualTo(2.0);
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-load-failure-time-total"))
+            .asInstanceOf(InstanceOfAssertFactories.DOUBLE)
+            .isEqualTo(0);
```
AnatolyPopov, line 106: Nit: should we preserve the same style and if `asInstanceOf` is not needed just do not add it? `isEqualTo(0.0)` would work just fine here as below. Or otherwise using `asInstanceOf` everywhere? WDYT?

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/DiskBasedChunkCacheMetricsTest.java
```diff
@@ -0,0 +1,116 @@
...
+        diskBasedChunkCache.getChunk(REMOTE_LOG_SEGMENT_METADATA, segmentManifest, 0);
+        diskBasedChunkCache.getChunk(REMOTE_LOG_SEGMENT_METADATA, segmentManifest, 0);
+
+        final var objectName = new ObjectName("aiven.kafka.server.tieredstorage.cache:type=chunk-cache");
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-hits-total"))
+            .isEqualTo(1.0);
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-misses-total"))
+            .isEqualTo(1.0);
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "cache-load-success-time-total"))
+            .asInstanceOf(InstanceOfAssertFactories.DOUBLE)
```
AnatolyPopov, line 98: Nit: static import?

AnatolyPopov (approved):
LGTM, just a couple of super minor comments but still mergeable.

jeqo:
Thanks @AnatolyPopov . Did some additional refactoring on the tests to also include in-mem in the mix as instrumentation happens on the abstract class.

AnatolyPopov (approved):
LGTM

PR #324, opened 2023-07-19
Provide keyring in configuration

ivanyu:
This commit makes config working on RSA keyrings instead of single key. Instead of a single pair of public and private files, it's needed to provide:
1. The list of key pairs IDs: `encryption.key.pairs`.
2. For each ID, the public and private file: `encryption.key.pairs.ID.public.key.file` and `encryption.key.pairs.ID.private.key.file`. (See e.g. how Kafka Connect configures transformations).
3. The "active" ID that will be used for encryption: `encryption.key.pair.id`.

See `RemoteStorageManagerConfigTest` for examples.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/KeyPairPaths.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/KeyPairPaths.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2023 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.aiven.kafka.tieredstorage;
+
+import java.nio.file.Path;
+import java.util.Objects;
+
+public class KeyPairPaths {
+    public final Path publicKey;
+    public final Path privateKey;
+
+    public KeyPairPaths(final Path publicKey, final Path privateKey) {
+        this.publicKey = Objects.requireNonNull(publicKey, "publicKey cannot be null");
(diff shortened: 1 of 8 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/KeyPairPaths.java
jeqo, line 1: Maybe move it to `security` package?

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
```diff
@@ -196,14 +180,107 @@ public class RemoteStorageManagerConfig extends AbstractConfig {
...
+                ENCRYPTION_KEY_PAIR_ID_DOC
+            );
+            configDef.define(
+                ENCRYPTION_KEY_PAIRS_CONFIG,
+                ConfigDef.Type.LIST,
+                ConfigDef.NO_DEFAULT_VALUE,
+                ConfigDef.Importance.HIGH,
+                ENCRYPTION_KEY_PAIRS_DOC
+            );
+            final EncryptionConfig encryptionConfig1 = new EncryptionConfig(configDef, props);
```
jeqo, line 228: nit:
```suggestion
            final EncryptionConfig config = new EncryptionConfig(configDef, props);
```

jeqo (commented):
Looking good! just a small comment and should be good to approve from my side.

core/src/main/java/io/aiven/kafka/tieredstorage/KeyPairPaths.java
ivanyu: I see it more as a part of the `config` package that doesn't exist. We can do this move, but I'd probably better do this in a follow-up PR to not clutter the review space.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerConfig.java
ivanyu: Thanks, sure! I just first had `encryptionConfig2` too. Renamed to `interimEncryptionConfig`.

core/src/main/java/io/aiven/kafka/tieredstorage/KeyPairPaths.java
jeqo: Sure, created https://github.com/aiven/tiered-storage-for-apache-kafka/issues/326 to follow up

jeqo (approved):

PR #325, opened 2023-07-19
feat(metrics): add cache metrics to chunk cache disk

jeqo:
This commits introduces a set of metrics for disk size, counting writes and delete rates and total.

Used by disk-based cache.

Depends on #323


jeqo:
Close as replaced by recent PRs merged.

PR #332, opened 2023-07-24
feat(metrics:cache): Add metric for estimated cache size

jeqo:
Cache size is not included on default Caffeine stats. Piggybacking on counter to record updates to estimated size.

Reference: https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineStatsCounter.java

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/metrics/CaffeineStatsCounter.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/metrics/CaffeineStatsCounter.java
@@ -36,6 +36,14 @@ import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import com.github.benmanes.caffeine.cache.stats.StatsCounter;
 
+/**
+ * Records cache metrics managed by Caffeine {@code Cache#stats}.
+ *
+ * <p>Note that this doesn't instrument the cache's size by default. Use
+ * {@link #registerSizeMetric(Supplier)} to do so after the cache has been built.
+ * Similar approach used by
+ * <a href="https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/cache/CaffeineStatsCounter.java">Micrometer</a>
+ */
 public class CaffeineStatsCounter implements StatsCounter {
 
     static final String CACHE_HITS = "cache-hits";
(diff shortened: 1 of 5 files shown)


AnatolyPopov:
@jeqo could you rebase this one?

jeqo:
@AnatolyPopov, sure. PR rebased and updated to latest metrics. Take a look, thanks!

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/CaffeineStatsCounter.java
```diff
@@ -135,11 +155,13 @@ private void initSensor(final String sensorName, final String metricName, final
     @Override
     public void recordHits(final int count) {
         cacheHitCount.add(count);
+        updateCacheSize();
     }
 
     @Override
     public void recordMisses(final int count) {
         cacheMissCount.add(count);
+        incrementCacheSize(); // increment by one when missing
```
AnatolyPopov, line 164: Could you elaborate why we are incrementing the size here?

jeqo: Sure. Initially, I started by only adding the size supplier on `recordHit`; but metrics will stay unmodified until there's a hit, which is inconvenient.
To solve this I added this hook to increment by one, as a "miss" should mean that the cache will increase.

Another alternative may be to call the size supplier here but haven't tested if the value is already updated or not.

AnatolyPopov: I trying to say a different thing, we are ourselves calling `recordMiss` and there we are not doing anything to size metric. At the same time this method is called by Caffeine and I'm not sure if we need to do anything to size here at all of if we even get here for ChunkCache at least since we are calling `compute` all the time.

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/CaffeineStatsCounter.java
```diff
@@ -219,4 +244,26 @@ public double measure(final MetricConfig config, final long now) {
...
+
+        SuppliedValue(final Supplier<Long> value) {
+            this.value = value;
+        }
+
+        @Override
+        public double measure(final MetricConfig config, final long now) {
+            return value.get();
+        }
+    }
```
AnatolyPopov, line 268: I don't think we need these 2 classes, we can have just have one and provide LongAdder::sum as value supplier. Also I would prefer to keep the name of the original class.

AnatolyPopov (changes requested):

PR #339, opened 2023-07-26
Distribute dependency version lines

ivanyu:
This allows to avoid constant rebasing when Dependabot kicks in.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -21,8 +21,11 @@ archivesBaseName = "core"
 
 ext {
     bouncyCastleVersion = "1.75"
+
     caffeineVersion = "3.1.7"
+
     zstdVersion = "1.5.5-5"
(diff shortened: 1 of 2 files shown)


giuseppelillo (approved):

core/build.gradle
```diff
@@ -21,8 +21,11 @@ archivesBaseName = "core"
 
 ext {
     bouncyCastleVersion = "1.75"
+
     caffeineVersion = "3.1.7"
+
     zstdVersion = "1.5.5-5"
+
     jacksonVersion = "2.15.2"
```
AnatolyPopov, line 29: Can we have a comment about it(since we can not enforce) to remember that there should be a new line after each version? And could you please move the rest of the versions to the variables to have the same style?

PR #351, opened 2023-08-11
feat: Adopt segment custom metadata on write

jeqo:
Segment custom metadata, introduced in [KIP-917], will be adopted by plugin to collect additional processing metadata for further usage.

Initial fields:
- total bytes: Including manifest and indexes sizes to log segment size.
- object key and prefix: for further usage on read path. 

Depends on #354

To do:

- [x] Using trunk snapshot to get access to latest APIs.
- [x] Add custom fields to configuration.
- [x] Test RSM with custom metadata.

[KIP-917]: https://cwiki.apache.org/confluence/display/KAFKA/KIP-917%3A+Additional+custom+metadata+for+remote+log+segment

### DIFF
--- a/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
+++ b/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
@@ -57,6 +57,8 @@ import io.aiven.kafka.tieredstorage.chunkmanager.cache.InMemoryChunkCache;
 import io.aiven.kafka.tieredstorage.manifest.index.ChunkIndex;
 import io.aiven.kafka.tieredstorage.manifest.serde.DataKeyDeserializer;
 import io.aiven.kafka.tieredstorage.manifest.serde.DataKeySerializer;
+import io.aiven.kafka.tieredstorage.metadata.SegmentCustomMetadataField;
+import io.aiven.kafka.tieredstorage.metadata.SegmentCustomMetadataSerde;
 import io.aiven.kafka.tieredstorage.security.AesEncryptionProvider;
 import io.aiven.kafka.tieredstorage.security.DataKeyAndAAD;
 import io.aiven.kafka.tieredstorage.security.EncryptedDataKey;
@@ -66,6 +68,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.github.luben.zstd.Zstd;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -107,6 +110,8 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
         "test/topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.log";
     static final String TARGET_MANIFEST_FILE =
         "test/topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.rsm-manifest";
+
+    static final SegmentCustomMetadataSerde CUSTOM_METADATA_SERDE = new SegmentCustomMetadataSerde();
 
     private static List<Arguments> provideEndToEnd() {
         final List<Arguments> result = new ArrayList<>();
@@ -194,7 +199,8 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
             "encryption.enabled", Boolean.toString(encryption),
             "chunk.cache.class", cacheClass,
             "chunk.cache.path", cacheDir.toString(),
-            "chunk.cache.size", Integer.toString(100 * 1024 * 1024)
+            "chunk.cache.size", Integer.toString(100 * 1024 * 1024),
+            "custom.metadata.fields.include", "REMOTE_SIZE,OBJECT_PREFIX,OBJECT_KEY"
         ));
         if (encryption) {
             config.put("encryption.key.pair.id", KEY_ENCRYPTION_KEY_ID);
@@ -211,8 +217,10 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
         final LogSegmentData logSegmentData = new LogSegmentData(
             logFilePath, offsetIndexFilePath, timeIndexFilePath, txnIndexPath,
             producerSnapshotFilePath, ByteBuffer.wrap(LEADER_EPOCH_INDEX_BYTES));
-        rsm.copyLogSegmentData(REMOTE_LOG_METADATA, logSegmentData);
+        final Optional<RemoteLogSegmentMetadata.CustomMetadata> customMetadata =
+            rsm.copyLogSegmentData(REMOTE_LOG_METADATA, logSegmentData);
 
+        checkCustomMetadata(customMetadata);
         checkFilesInTargetDirectory(hasTxnIndex);
         checkManifest(chunkSize, compression, encryption);
         if (encryption) {
@@ -221,6 +229,21 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
         checkIndexContents(hasTxnIndex);
         checkFetching(chunkSize);
         checkDeletion();
+    }
+
+    private void checkCustomMetadata(final Optional<RemoteLogSegmentMetadata.CustomMetadata> customMetadata) {
+        assertThat(customMetadata).isPresent();
+        final var fields = CUSTOM_METADATA_SERDE.deserialize(customMetadata.get().value());
+        assertThat(fields).hasSize(3);
+        assertThat(fields.get(SegmentCustomMetadataField.REMOTE_SIZE.index()))
+            .asInstanceOf(InstanceOfAssertFactories.LONG)
+            .isGreaterThan(0);
+        assertThat(fields.get(SegmentCustomMetadataField.OBJECT_KEY.index()))
+            .asInstanceOf(InstanceOfAssertFactories.STRING)
+            .isNotEmpty();
+        assertThat(fields.get(SegmentCustomMetadataField.OBJECT_PREFIX.index()))
+            .asInstanceOf(InstanceOfAssertFactories.STRING)
+            .isNotEmpty();
     }
 
     private void checkFilesInTargetDirectory(final boolean hasTxnIndex) {
@@ -234,9 +257,8 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
         if (hasTxnIndex) {
             expectedFiles.add("topic-AAAAAAAAAAAAAAAAAAAAAQ/7/00000000000000000023-AAAAAAAAAAAAAAAAAAAAAA.txnindex");
         }
-        expectedFiles.forEach(s -> {
-            assertThat(targetDir).isDirectoryRecursivelyContaining(path -> path.toString().endsWith(s));
-        });
+        expectedFiles.forEach(s ->
+            assertThat(targetDir).isDirectoryRecursivelyContaining(path -> path.toString().endsWith(s)));
     }
 
     private void checkManifest(final int chunkSize,
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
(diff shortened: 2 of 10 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataField.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.kafka.common.protocol.types.Field;
+import org.apache.kafka.common.protocol.types.Field.TaggedFieldsSection;
+import org.apache.kafka.common.protocol.types.Schema;
+import org.apache.kafka.common.protocol.types.Type;
+
+import io.aiven.kafka.tieredstorage.ObjectKey;
+
+// ordinal index define values on custom metadata fields and cannot be changed without breaking compatibility.
+public enum SegmentCustomMetadataField {
+    TOTAL_SIZE(new Field("total_size", Type.VARLONG), SegmentCustomMetadataBuilder::totalSize),
```
ivanyu, line 31: `remote_size` maybe (here and in other places), WDYT?

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataField.java
```diff
@@ -0,0 +1,57 @@
...
+    OBJECT_PREFIX(new Field("object_prefix", Type.COMPACT_STRING), b -> b.objectKey.prefix()),
+    OBJECT_KEY(new Field("object_key", Type.STRING), b -> ObjectKey.mainPath(b.segmentMetadata));
```
ivanyu, line 33: Any reason not to use `COMPACT_STRING` here as well?

core/src/test/java/io/aiven/kafka/tieredstorage/config/RemoteStorageManagerConfigTest.java
```diff
@@ -307,4 +309,33 @@ void invalidCompressionConfig() {
             .isInstanceOf(ConfigException.class)
             .hasMessage("compression.enabled must be enabled if compression.heuristic.enabled is");
     }
+
+    @Test
+    void validCustomMetadataFieldsLowercase() {
```
ivanyu, line 314:
```suggestion
    void validCustomMetadataFieldsUppercase() {
```
I guess?

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+        this.fields = fields;
+        this.objectKey = objectKey;
+        this.segmentMetadata = segmentMetadata;
+        this.uploadResults = new HashMap<>(ObjectKey.Suffix.values().length);
+    }
+
+    public SegmentCustomMetadataBuilder addUploadResult(final ObjectKey.Suffix suffix,
+                                                        final String keyPath,
+                                                        final long bytes) {
+        uploadResults.put(suffix, new UploadResult(keyPath, bytes));
```
ivanyu, line 47: Maybe it's worth checking that the key doesn't exist, just as a sanity check

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
```diff
@@ -223,6 +229,12 @@ void endToEnd(
         checkDeletion();
     }
 
+    private void checkCustomMetadata(final Optional<RemoteLogSegmentMetadata.CustomMetadata> customMetadata) {
+        assertThat(customMetadata).isPresent();
+        final var fields = CUSTOM_METADATA_SERDE.deserialize(customMetadata.get().value());
+        assertThat(fields).hasSize(3);
```
ivanyu, line 235: Maybe also check at least size > 0 and the values of prefix and key here?

ivanyu:
Apart from the comments I left, all LGTM. @jeqo how do you think we should deal with this code? Merge or keep the branch until 3.5 is released?

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataField.java
jeqo: Agree, applying the change.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataField.java
jeqo: Not really, forgot to align these. Should be fixed now 👍🏽

core/src/test/java/io/aiven/kafka/tieredstorage/config/RemoteStorageManagerConfigTest.java
jeqo: Of course 😓

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
jeqo: Do you mean like checking with the storage backend that the file is not missing? or something else?

jeqo:
Thanks @ivanyu! Let's fix #354 and #353 before merging this one.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
ivanyu: Like 
```java
if (uploadResults.containsKey(suffix)) {
    throw ...
}
```

jeqo: Of course. Good idea!

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+        uploadResults.put(suffix, new UploadResult(keyPath, bytes));
+        return this;
+    }
+
+    public long totalSize() {
+        return uploadResults.values().stream().mapToLong(value -> value.bytes).sum();
+    }
+
+    public Map<Integer, Object> build() {
```
ivanyu, line 58: Tagged fields [require](https://github.com/apache/kafka/blob/f44ee4fab7ef7adf715ecf2b96defa5cc8311949/clients/src/main/java/org/apache/kafka/common/protocol/types/TaggedFields.java#L153) `NavigableMap`, we should use it as the type instead of just `Map`. I know we pass `TreeMap` anyway, but it makes sense to ensure this on the type level.

ivanyu (changes requested):
LGTM apart from one request above

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
jeqo: Agree. Applying suggestion.

ivanyu (dismissed):

core/src/main/java/io/aiven/kafka/tieredstorage/config/RemoteStorageManagerConfig.java
```diff
@@ -178,6 +187,14 @@ public class RemoteStorageManagerConfig extends AbstractConfig {
                     Sensor.RecordingLevel.TRACE.toString()),
                 ConfigDef.Importance.LOW,
                 METRICS_RECORDING_LEVEL_DOC);
+
+        CONFIG
```
AnatolyPopov, line 191: Should we keep the formatting consistent?

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+    public NavigableMap<Integer, Object> build() {
+        if (fields.isEmpty()) {
+            return new TreeMap<>();
+        }
+        final TreeMap<Integer, Object> taggedFields = new TreeMap<>();
+        fields.forEach(field -> taggedFields.put(field.ordinal(), field.valueProvider.apply(this)));
+        return taggedFields;
+    }
+
+    private static class UploadResult {
```
AnatolyPopov, line 68: Can you elaborate why do we need this class if we are not using it externally?

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+        uploadResults.put(suffix, new UploadResult(keyPath, bytes));
+        return this;
+    }
+
+    public long totalSize() {
+        return uploadResults.values().stream().mapToLong(value -> value.bytes).sum();
+    }
+
+    public NavigableMap<Integer, Object> build() {
+        if (fields.isEmpty()) {
```
AnatolyPopov, line 60: If we put nothing in the map then it's empty right? The check seems to be redundant

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.TreeMap;
+
+import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentMetadata;
+
+import io.aiven.kafka.tieredstorage.ObjectKey;
+
+public class SegmentCustomMetadataBuilder {
+    final ObjectKey objectKey;
+    final RemoteLogSegmentMetadata segmentMetadata;
+    final Map<ObjectKey.Suffix, UploadResult> uploadResults;
```
AnatolyPopov, line 32: Together with the question about `UploadResult` necessity, can this be just `Map<ObjectKey.Suffix, Long>`?

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -265,7 +292,8 @@ boolean requiresCompression(final LogSegmentData logSegmentData) {
     }
 
     private void uploadSegmentLog(final RemoteLogSegmentMetadata remoteLogSegmentMetadata,
-                                  final TransformFinisher transformFinisher)
+                                  final TransformFinisher transformFinisher,
+                                  final SegmentCustomMetadataBuilder customMetadataBuilder)
```
AnatolyPopov, line 296: Here and in other places: should we add more arguments or would it better to return the size instead?

AnatolyPopov (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/config/RemoteStorageManagerConfig.java
jeqo: Agree, fixing it on this and above metric configs.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    final Set<SegmentCustomMetadataField> fields;
+
+    public SegmentCustomMetadataBuilder(final Set<SegmentCustomMetadataField> fields,
+                                        final ObjectKey objectKey,
+                                        final RemoteLogSegmentMetadata segmentMetadata) {
+        this.fields = fields;
+        this.objectKey = objectKey;
+        this.segmentMetadata = segmentMetadata;
+        this.uploadResults = new HashMap<>(ObjectKey.Suffix.values().length);
```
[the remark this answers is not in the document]
jeqo: Good catch, adding this type.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
jeqo: Was initially planning to use it for other metadata fields including paths. But as you spotted, no need for it, and Long is enough.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
jeqo: Good catch. Removing it.

core/src/main/java/io/aiven/kafka/tieredstorage/metadata/SegmentCustomMetadataBuilder.java
jeqo: Moving to long.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: There may be some refactoring possible, though I'd suggest doing it in a separate PR.

AnatolyPopov (approved):
LGTM

PR #353, opened 2023-08-15
chore(ci): Update CI workflows

jeqo:
Currently, all workflows run on every PR, causing duplicated executions. This PR fixes this, and separate tests into different steps to easier troubleshooting. 
Also, includes building docker image.


.github/workflows/pull_request_workflow.yml
```diff
@@ -20,5 +20,12 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
           distribution: temurin
+
       - name: Build with Gradle
-        run: ./gradlew build distTar
+        run: ./gradlew build distTar -x test
+
+      - name: Run integration tests
+        run: ./gradlew integrationTest
```
AnatolyPopov, line 28: So we are skipping the unit tests entirely, right?

.github/workflows/pull_request_workflow.yml
```diff
@@ -20,5 +20,12 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
           distribution: temurin
+
       - name: Build with Gradle
-        run: ./gradlew build distTar
+        run: ./gradlew build distTar -x test
```
AnatolyPopov, line 25: And do we continue running the integrationTests here despite the next stage?

.github/workflows/pull_request_workflow.yml
jeqo: My bad. Meant to disable integrationTest on previous state. Should be fixed now.

.github/workflows/main_push_workflow.yml
```diff
@@ -2,7 +2,7 @@
 name: Main checks after push
 on:
   push:
-    branches: [ '**' ]
+    branches: [ 'main' ]
```
AnatolyPopov, line 5: So we are not running the pipeline on push to branches anymore?

jeqo: Not until PRs are created. As far as I remember we opened to all branches to build workflows while working on the new implementation. Also, this aligns with other repos: https://github.com/Aiven-Open/s3-connector-for-apache-kafka/blob/main/.github/workflows/main_push_and_pull_request_workflow.yml

Let me know if I'm missing something, though.

AnatolyPopov: Ah, yeah, I remembered. But should we actually have 2 workflows or 1 with 2 conditions?

jeqo: We could actually. Was considering keeping them separate to publish images only (TODO) only on the main branch; but could be also done on PRs with a different tag 🤔 

If we prefer this, I'm happy to merge them

AnatolyPopov (approved):

PR #328, opened 2023-07-21
feat(e2e): move e2e tests

jeqo:
See commits for details on changes.

Depends on 
- #364

PR #354, opened 2023-08-15
chore(build): Switch to upstream binaries

jeqo:
Update binaries to latest upstream build, to use newer APIs introduced in KIP-917.

Note, this will temporarily break the demos as base docker image will need to be based on upstream as well; otherwise plugin is incompatible and will break at runtime.

Depends on:

- https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/355
- https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/353
- #328


jeqo:
Requires https://github.com/aiven/kafka/pull/36 to publish images to DockerHub

docker/Dockerfile
```diff
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ##
-FROM aivenoy/kafka:3.3-2022-10-06-tiered-storage-1
+# TODO: fix once `aivenoy/` image is available on docker hub
+FROM jeqo/kafka:kips_KIP-405_2023-08-18
```
jeqo, line 17: just for testing purposes, I used an image from my own repo, and CI is passing. We can switch to `aivenoy` image when available.

ivanyu (commented):
LGTM, but I wonder why the demos were expanded to run two brokers?

jeqo:
> wonder why the demos were expanded to run two brokers?

good catch. I expanded it to try to reproduce some issues, but I can leave it out and add it in a different PR if needed.

ivanyu (approved):

AnatolyPopov (approved):
LGTM

PR #356, opened 2023-08-16
refactor: Pass key path instead of log metadata + object key

jeqo:
To reduce coupling, and define path on RSM instead of down the chunk manager. A nice side effect is that dependencies on metadata get reduced to only RSM (where it belongs). Object Key also ends up being used only by the RSM

This changes are required by following PRs, as path to a segment can be based on the custom metadata (available on RSM). To avoid having to send another field all the way down to transformers, etc. with this change, all can be solved at the RSM level (as it's solved for indexes and manifest file.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkKey.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkKey.java
@@ -16,20 +16,17 @@
 
 package io.aiven.kafka.tieredstorage.chunkmanager;
 
+import java.nio.file.Path;
 import java.util.Objects;
 
-import org.apache.kafka.common.Uuid;
-import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentId;
-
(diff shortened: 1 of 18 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/ChunkKey.java
```diff
@@ -16,20 +16,17 @@
...
 
+import java.nio.file.Path;
 import java.util.Objects;
 
-import org.apache.kafka.common.Uuid;
-import org.apache.kafka.server.log.remote.storage.RemoteLogSegmentId;
-
 public class ChunkKey {
-    public final Uuid uuid;
+    public final Path segmentFileName;
```
AnatolyPopov, line 23: Should we change this to be just `String` instead? Since we have in memory cache also and from it's perspective path looks weird same as file name unless we say specifically that it's an s3 filename somewhere.

AnatolyPopov (approved):
LGTM

PR #363, opened 2023-08-21
fix(demo): improve Makefile commands

jeqo:
Changes:

- Add replication factor variable
- Disable time-based retention when size-based topics are created.

### DIFF
--- a/demo/Makefile
+++ b/demo/Makefile
@@ -35,6 +35,7 @@ create_topic_ts_by_size:
 		--bootstrap-server $(bootstrap_server) \
 		--create \
 		--config remote.storage.enable=true \
+		--config retention.ms=-1 \
 		--config segment.bytes=$(segment) \
 		--config retention.bytes=$(retention_bytes) \
 		--config local.retention.bytes=$(local_retention_bytes) \
@@ -62,6 +63,7 @@ create_topic_no_ts_by_size:
 		kafka-topics \
 		--bootstrap-server $(bootstrap_server) \
 		--create \
+		--config retention.ms=-1 \
 		--config segment.bytes=$(segment) \
 		--config retention.bytes=$(retention_bytes) \
 		--partitions $(partitions) \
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -132,9 +136,17 @@ subprojects {
         testRuntimeOnly "org.slf4j:slf4j-log4j12:$slf4jVersion"
     }
 
-    tasks.named('test') {
+    test {
```
AnatolyPopov, line 139: Please read gradle guidlines https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_pitfalls
This should actually be applied everywhere and I was trying to do it before but maybe I missed some places.

AnatolyPopov (changes requested):

build.gradle
jeqo: Thanks, @AnatolyPopov! Wasn't aware of this. Gradle can be very tricky sometimes.

Though, this is probably referring to changes from #328. Applied the fixes there. 

I have removed the dependency on that PR as changes here can be applied individually. Have a look

demo/Makefile
```diff
@@ -59,9 +63,11 @@ create_topic_no_ts_by_size:
 		kafka-topics \
 		--bootstrap-server $(bootstrap_server) \
 		--create \
+		--config retention.ms=-1 \
 		--config segment.bytes=$(segment) \
```
AnatolyPopov, line 67: What does `no_ts_by_size` mean? Is it by size or by time? From the code I can guess that it's by size, but the name is weird in that case. Same in other similarly named tasks

AnatolyPopov (changes requested):

demo/Makefile
jeqo: sure, applying suggestions. have a look.

AnatolyPopov (approved):
LGTM

PR #364, opened 2023-08-22
fix: handle jackson dependency conflict between core module and s3

### LINKED ISSUE #254: Exclude Jackson from distribution
Jackson is present in Kafka, so it shouldn't be included in the plugin distribution.

jeqo:
Fix #254 

The dependency issue reported on #254 is not between the version used by the plugin and the broker (as stated in the comments, it's handled by a custom class loader on the broker side); but by different versions used on the core and s3 module.

This PR includes the build changes to handle this exclusion, and the exception handling to capture this. Also some additional logging improvements.

### DIFF
--- a/storage/s3/build.gradle
+++ b/storage/s3/build.gradle
@@ -27,7 +27,14 @@ ext {
 dependencies {
     implementation project(":storage:core")
 
-    implementation "com.amazonaws:aws-java-sdk-s3:$awsSdkVersion"
+    // Take out jackson libraries out to avoid collision with core jackson dependencies
+    implementation ("com.amazonaws:aws-java-sdk-s3:$awsSdkVersion") {
+        exclude group: "com.fasterxml.jackson.core"
+        exclude group: "com.fasterxml.jackson.dataformat"
+    }
+    // replacing with version used by core
+    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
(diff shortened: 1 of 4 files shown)


ivanyu (approved):

storage/s3/build.gradle
```diff
@@ -27,7 +27,14 @@ ext {
 dependencies {
     implementation project(":storage:core")
 
-    implementation "com.amazonaws:aws-java-sdk-s3:$awsSdkVersion"
+    // Take out jackson libraries out to avoid collision with core jackson dependencies
+    implementation ("com.amazonaws:aws-java-sdk-s3:$awsSdkVersion") {
+        exclude group: "com.fasterxml.jackson.core"
+        exclude group: "com.fasterxml.jackson.dataformat"
```
AnatolyPopov, line 33: We still can merge but I wonder if there is a proper way in Gradle to enforce versions of transitive dependencies similar to maven dependencyManagement section. That would allow to solve this kind of issues once and for all. And in this case if there will be more dependencies bringing jackson as transitive we will have to add more exclusions.

AnatolyPopov (approved):

storage/s3/build.gradle
jeqo: Agree, we should look for some way to do this from Gradle. Will have a look.

jeqo: I tried to add a custom task to handle this validation. Have a look, thank!

ivanyu (approved):

PR #365, opened 2023-08-31
Migration to AWS SDK v2

AnatolyPopov:


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -209,12 +223,25 @@ public int uploadPartSize() {
         return getInt(S3_MULTIPART_UPLOAD_PART_SIZE_CONFIG);
     }
 
+    private URI s3ServiceEndpoint() {
+        final String url = getString(S3_ENDPOINT_URL_CONFIG);
+        if (url != null) {
+            try {
+                return new URI(url);
+            } catch (final URISyntaxException e) {
+                throw new ConfigException(S3_ENDPOINT_URL_CONFIG, url, "Must be a valid URI");
```
ivanyu, line 232: I think this is worth adding a validator, but can be done separately.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -153,47 +156,58 @@ && getPassword(AWS_ACCESS_KEY_ID_CONFIG) != null) {
         }
     }
 
-    AmazonS3 s3Client() {
-        final AmazonS3ClientBuilder s3ClientBuilder = AmazonS3ClientBuilder.standard();
+    S3Client s3Client() {
+        final S3ClientBuilder s3ClientBuilder = S3Client.builder();
         final String s3ServiceEndpoint = getString(S3_ENDPOINT_URL_CONFIG);
         final String region = getString(S3_REGION_CONFIG);
         if (Objects.isNull(s3ServiceEndpoint)) {
```
ivanyu, line 163:
```suggestion
        if (Objects.isNull(s3ServiceEndpoint())) {
```
I guess? Then the variable can be removed

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -209,12 +223,25 @@ public int uploadPartSize() {
...
+        }
+    }
+
     private static class RegionValidator implements ConfigDef.Validator {
         @Override
         public void ensureValid(final String name, final Object value) {
             final String regionStr = (String) value;
             try {
-                Regions.fromName(regionStr);
+                Region.of(regionStr);
```
ivanyu, line 244: I think this no longer works for validation, because `Region.of` doesn't seem to throw on an unknown region. OTOH, we can also accept everything the user provides as the region :man_shrugging: (i.e. drop the validator).

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
```diff
@@ -16,153 +16,133 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 class S3StorageConfigTest {
+    private static final String BUCKET_NAME = "b1";
+    private static final Region TEST_REGION = Region.US_EAST_2;
+    private static final String MINIO_URL = "http://minio";
 
     // Test scenarios
     // - Minimal/Full configs
```
ivanyu, line 41: I think the full config test wasn't here before neither is here now, so probably
```suggestion
    // - Minimal config
```

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -76,35 +68,32 @@ private Sensor createSensor(final String name) {
...
-        }
-
-        // Try to arrange them by likelihood, at least GetObjectRequest should be first.
-        if (GetObjectRequest.class.getSimpleName().equals(requestType)) {
+    public void publish(final MetricCollection metricCollection) {
+        if (metricCollection.metricValues(CoreMetric.OPERATION_NAME).contains("GetObject")) {
             this.getObjectRequests.record();
-        } else if (UploadPartRequest.class.getSimpleName().equals(requestType)) {
+        }
+        if (metricCollection.metricValues(CoreMetric.OPERATION_NAME).contains("UploadPart")) {
```
ivanyu, line 75: I think `metricCollection.metricValues(CoreMetric.OPERATION_NAME)` should be in a variable (it does some iteration inside, which maybe not super optimal).

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -76,35 +68,32 @@ private Sensor createSensor(final String name) {
...
             this.uploadPartRequests.record();
```
ivanyu, line 76: Can the collection contain more than one event for an operation?

ivanyu:
Apart from what's said, LGTM

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
AnatolyPopov: Fixed

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
AnatolyPopov: Agree, dropping the validator since we should also allow custom regions

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
AnatolyPopov: Fixed

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
AnatolyPopov: Yeah, lets see later if we need it, after applying one of your other comments the validator does not seem to be necessary since it will anyways lead to manual call to getter or constructor of URI.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
AnatolyPopov: hm, actually not sure but the metrics test is passing...

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -76,35 +68,35 @@ private Sensor createSensor(final String name) {
...
+            if ("DeleteObject".equals(metricValue)) {
+                this.deleteObjectRequests.record();
+            }
+            if ("AbortMultipartUpload".equals(metricValue)) {
+                this.abortMultipartUploadRequests.record();
+            }
         }
+    }
 
-        // Try to arrange them by likelihood, at least GetObjectRequest should be first.
```
ivanyu, line 93: Probably makes sense to preserve this commit

ivanyu (changes requested):
LGTM, one small request only

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
AnatolyPopov: You mean comment, right?

ivanyu: Yeah, sorry :)

ivanyu (approved):

Issue #158, opened 2023-04-06
[New implementation] Consider adding more metadata to manifest to make remote segments self-sufficient


ivanyu:
First think if we need this at all. To be able to restore topic from the remote storage in case of disaster is good.

The metadata in `RemoteLogSegmentMetadata`:
- `RemoteLogSegmentId` - encoded in the path;
- `startOffset` - encoded in the parth;
- `endOffset`;
- `maxTimestampMs`;
- `segmentLeaderEpochs`;
- `segmentSizeInBytes`;
- `RemoteLogSegmentState` - should always be `COPY_SEGMENT_STARTED` when `copyLogSegmentData` is called.

PR #377, opened 2023-09-07
Store remote log segment metadata

ivanyu:
To make the segments more self-contained, this commits adds the remote log segment metadata in the manifest. However, it doesn't serialize it back (we don't need it).

Closes https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/issues/158


jeqo (approved):
LGTM

PR #383, opened 2023-09-13
Add option to disable SSL cert verification in the S3 client

eliax1996:
It adds an option to disable the ssl cert verification, it's useful in testing scenario when it's hard/complex to install the generated root certificate into the kafka hosting machine.

### DIFF
--- a/storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
+++ b/storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
@@ -51,6 +51,7 @@ class S3StorageConfigTest {
         assertThat(config.credentialsProvider()).isNull();
         assertThat(config.pathStyleAccessEnabled()).isNull();
         assertThat(config.uploadPartSize()).isEqualTo(S3_MULTIPART_UPLOAD_PART_SIZE_DEFAULT);
+        assertThat(config.disableAwsCertChecking()).isEqualTo(false);
 
         verifyClientConfiguration(config.s3Client(), null);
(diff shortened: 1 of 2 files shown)


storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
```diff
@@ -108,7 +109,8 @@ void configWithStaticCredentials() {
             "s3.region", region.id(),
             "s3.endpoint.url", MINIO_URL,
             "aws.access.key.id", username,
-            "aws.secret.access.key", password);
+            "aws.secret.access.key", password,
+            S3StorageConfig.DISABLE_AWS_CERT_CHECKING_CONFIG, "true");
```
jeqo, line 113: nit: could we use a string instead of a field as other configs? We do this to enforce config names do not change inadvertently.

jeqo (commented):
@eliax1996 thanks! New config and docs look good to me. Left a minor comment on the unit test.

Wondering if there's a way to expand the S3 integration test to validate this new configuration. Or maybe adapt the demo with MinIO to have a way to validate? wdyt?

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
eliax1996: sure

jeqo (approved):
LGTM, thanks!

We can update demo in another PR.

AnatolyPopov (approved):

PR #384, opened 2023-09-13
fixed Makefile, perl isn't supported in grep for mac

eliax1996:
I've replaced the perl syntax to keep only a part of the regex with a grep followed by a cat to remove the unwanted chars

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ##
-VERSION := $(shell grep -oP 'version=\K[^[:space:]]+' gradle.properties)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ##
-VERSION := $(shell grep -oP 'version=\K[^[:space:]]+' gradle.properties)
+VERSION := $(shell grep -o '^version=.*-SNAPSHOT' gradle.properties | cut -c9-)
```
AnatolyPopov, line 16: Is it looking for `SNAPHSOT` in version? How it will work for a release version?

eliax1996: fair, let me fix it 😄

eliax1996: what about now?

jeqo (approved):

PR #394, opened 2023-09-21
Prefetching a configurable amount of data.

AnatolyPopov:


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -33,6 +33,12 @@ public class ChunkCacheConfig extends AbstractConfig {
             + "where \"-1\" represents infinite retention";
     private static final long DEFAULT_CACHE_RETENTION_MS = 600_000;
 
+    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetching.bytes";
+    private static final String CACHE_PREFETCHING_SIZE_DOC =
+        "The amount of data that should be eagerly prefetched and cached";
+
+    private static final int CACHE_PREFETCHING_SIZE_DEFAULT = 0; //TODO find out what it should be
```
jeqo, line 40: Thinking about defaults, we can target something like 30MB.
The goal is to target 2x the max bytes per partition. If we suggest max fetch bytes per partition to be 15MB, using 30MB/s here will help use to have _the following_ fetch request prepared locally. 

More than 30MB may be too much in term of network allocation.

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -33,6 +33,12 @@ public class ChunkCacheConfig extends AbstractConfig {
             + "where \"-1\" represents infinite retention";
     private static final long DEFAULT_CACHE_RETENTION_MS = 600_000;
 
+    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetching.bytes";
```
jeqo, line 36: maybe `prefetch.size` based on number of chunks will be easier to configure. 
Even though an implementation detail, chunk size is a pretty core concept on the plugin  (e.g. config will be named `chunk.cache.prefetch.size`)

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCache.java
```diff
@@ -138,4 +142,29 @@ protected AsyncCache<ChunkKey, T> buildCache(final ChunkCacheConfig config) {
         statsCounter.registerSizeMetric(cache.synchronous()::estimatedSize);
         return cache;
     }
+
+    public void startPrefetching(final String segmentKey,
```
jeqo, line 146: Wondering if this could be refactor to receive a set of chunk ids to prefetch instead of calculate based on range, and move the calculation to the FetchChunkEnumeration constructor where start and end chunk id are already calculated.

If we switch the config to be by number of chunks instead of bytes, then we can do the prefetching directly here instead of the initial prep. wdyt?

jeqo (commented):
Thanks, @AnatolyPopov .

This looks promising. I just have some comments about config definition, and some suggestions to refactor the prefetching logic. 

Overall, the approach looks sound to me: Pre-fetching once per fetch request. 
For this to be effective, we should consider the prefetched size to be at least higher than the chunk size (else no chunk will be cached) and, better, to be equal or larger than the max fetch bytes per partition, so at least all the get requests to e.g. S3 will be done at once. If double, we can expect the next fetch request to be sourced directly from local cache.

Also, cache size will be affected by this. If 30MB is prefetched per partition, then a broker to prefetch from 10 partitions concurrently will require at least 300MB to cache all required chunks.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -452,7 +453,9 @@ public InputStream fetchLogSegment(final RemoteLogSegmentMetadata remoteLogSegme
 
             final var suffix = ObjectKeyFactory.Suffix.LOG;
             final var segmentKey = objectKey(remoteLogSegmentMetadata, suffix);
-
+            if (chunkManager instanceof ChunkCache) {
+                ((ChunkCache<?>) chunkManager).startPrefetching(segmentKey, segmentManifest, range.from);
+            }
```
ivanyu, line 458: Any reason this is happening here and not in `ChunkCache#getChunk`? Looks like a leaky abstraction.

AnatolyPopov: You are right, fixed.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/ChunkIndex.java
```diff
@@ -49,4 +50,6 @@ public interface ChunkIndex {
      * Returns all chunks in the index.
      */
     List<Chunk> chunks();
+
+    List<Chunk> listChunksForRange(BytesRange bytesRange);
```
ivanyu, line 54: Not used?

ivanyu (changes requested):
Looks good, a couple of minor things

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/ChunkIndex.java
AnatolyPopov: Removed

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfigTest.java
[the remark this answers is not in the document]
AnatolyPopov: Added

ivanyu (approved):

PR #397, opened 2023-09-26
feat: Add S3 latency metrics

jeqo:
For better monitoring, latency metrics will help to understand how long are we waiting on the plugin side for a response from S3.

See commits for more details, a small refactor on the testing is included separately.

### DIFF
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
@@ -27,10 +27,14 @@ import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
 import org.apache.kafka.common.metrics.MetricsReporter;
 import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.common.metrics.stats.Avg;
 import org.apache.kafka.common.metrics.stats.CumulativeCount;
+import org.apache.kafka.common.metrics.stats.Max;
 import org.apache.kafka.common.metrics.stats.Rate;
 import org.apache.kafka.common.utils.Time;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.core.metrics.CoreMetric;
 import software.amazon.awssdk.metrics.MetricCollection;
 import software.amazon.awssdk.metrics.MetricPublisher;
@@ -42,10 +46,13 @@ import static software.amazon.awssdk.core.internal.metrics.SdkErrorType.SERVER_E
 import static software.amazon.awssdk.core.internal.metrics.SdkErrorType.THROTTLING;
 
 class MetricCollector implements MetricPublisher {
+    private static final Logger log = LoggerFactory.getLogger(MetricCollector.class);
+
     private final org.apache.kafka.common.metrics.Metrics metrics;
 
     private static final String METRIC_GROUP = "s3-metrics";
     private final Map<String, Sensor> requestMetrics = new HashMap<>();
+    private final Map<String, Sensor> latencyMetrics = new HashMap<>();
     private final Map<String, Sensor> errorMetrics = new HashMap<>();
 
     MetricCollector() {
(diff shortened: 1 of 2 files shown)


storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -55,41 +58,67 @@ class MetricCollector implements MetricPublisher {
...
+
     @Override
     public void publish(final MetricCollection metricCollection) {
         final List<String> metricValues = metricCollection.metricValues(CoreMetric.OPERATION_NAME);
-        for (final String metricValue : metricValues) {
-            final var sensor = requestMetrics.get(metricValue);
-            if (sensor != null) {
-                sensor.record();
+        // metrics are reported per request, so 1 value can be assumed.
+        if (metricValues.size() == 1) {
```
AnatolyPopov, line 101: The for loop was introduced intentionally after the discussion with @ivanyu and I guess it will not affect anything if what you specified in the comment is true, however both me and @ivanyu were not sure if that's the case. So could you elaborate how did you find out that the size will always be 1?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -55,41 +58,67 @@ class MetricCollector implements MetricPublisher {
...
+            final var metricValue = metricValues.get(0);
+            final var requests = requestMetrics.get(metricValue);
+            if (requests != null) {
+                requests.record();
+            }
+
+            final var durations = metricCollection.metricValues(CoreMetric.API_CALL_DURATION);
+            if (durations.size() == 1) {
```
AnatolyPopov, line 109: Does this need to be inside the previous if? And same question regarding number of values.

AnatolyPopov (changes requested):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
jeqo: Yeah, it's to get the right sensor, as the metric for latency only contains the duration. As the metric collection is correlated, then I'm using this value to update metrics properly.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
jeqo: API is not clear about it, but docs mention the following: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics-list.html 

> Metrics collected with each request

So, my interpretation (and while testing this) is that each instance of metric collection corresponds to a single API call.

AnatolyPopov: Ok, I looked into the SDK code and seems you are right :+1:

AnatolyPopov (dismissed):
LGTM

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
ivanyu: But OTOH what harm would it do it it remains a loop? It'd be more future-proof and we wouldn't expect this internal behavior to change in a casual dependency upgrade

AnatolyPopov: The thing is that we will not be able to understand to what request type the latency metric belongs if there will be multiple values AFAIU. Otherwise I would agree.

ivanyu: Oh, I see. That's very unfortunate. Well, let's rely on our tests as much as we can then

ivanyu (approved):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
ivanyu, line 101: Hm, I'm thinking that maybe it's worth logging a warning if the size is different from 1

ivanyu (changes requested):

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
AnatolyPopov: I was thinking the same actually, so yeah, lets have it.

jeqo: Make sense. Adding it now.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java
```diff
@@ -55,41 +62,69 @@ class MetricCollector implements MetricPublisher {
...
```
ivanyu, line 113: I think we need the same warning in the `else` here

ivanyu (changes requested):
One more comment, but otherwise LGTM

ivanyu (approved):

PR #401, opened 2023-09-27
 feat: add metrics for chunk transform/detransform latency 

jeqo:
Depends on #400


core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -55,6 +55,7 @@
 import io.aiven.kafka.tieredstorage.metadata.SegmentCustomMetadataBuilder;
 import io.aiven.kafka.tieredstorage.metadata.SegmentCustomMetadataField;
 import io.aiven.kafka.tieredstorage.metadata.SegmentCustomMetadataSerde;
+import io.aiven.kafka.tieredstorage.metrics.MeteredInputStream;
```
ivanyu, line 58: I guess this one is not added yet?

jeqo: 🤦🏽 of course, adding it now.

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/MeteredInputStream.java
```diff
@@ -0,0 +1,54 @@
...
+
+package io.aiven.kafka.tieredstorage.metrics;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.function.Consumer;
+
+import org.apache.kafka.common.utils.Time;
+
+public class MeteredInputStream extends InputStream {
```
ivanyu, line 25: There's a problem with this implementation: direct inheritance from `InputStream` and overriding only `read()` will result in other read methods being implemented through `read()`, which is very inefficient, because it's by-byte.

I looked at this problem some time ago and here's my draft: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/tree/ivanyu/time-to-first-byte. Using `FilterInputStream` allows to override fewer methods.

ivanyu (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/metrics/MeteredInputStream.java
jeqo: Thanks for the pointer! Yes, at the end other implementations use the read method, but as you mentioned it could be inefficient. Will revisit this then.

PR #403, opened 2023-09-28
refactor: fetch and catch parts of segment instead of chunks

jeqo:
To decouple fetching rate from transformation, a new concept of Part is introduced. The goal is to download larger ranges (parts) and transform smaller chunks.
Caching also gets the benefit of having less fragmented bits of data. By having fetch parts instead of chunks, chunks will be collocated on the same file (in the case of disk-based cache) and could be fetched together faster. With the current approach, multiple reads to separate files are required.

It also includes a renaming from ChunkManager/Cache into FetchManager/Cache. Beware the 3rd commit will be large but only includes renaming. Main changes are on the second commit. Apart from first commit, others will be squashed if approved.


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/DefaultChunkManager.java
```diff
@@ -46,13 +45,13 @@ public DefaultChunkManager(final ObjectFetcher fetcher, final AesEncryptionProvi
...
+    @Override
+    public InputStream chunksContent(final String objectKeyPath,
+                                     final SegmentManifest manifest,
+                                     final Part part) throws StorageBackendException {
+        final InputStream chunkContent = fetcher.fetch(objectKeyPath, part.range);
 
-        final InputStream chunkContent = fetcher.fetch(objectKeyPath, chunk.range());
-
-        DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(chunkContent, List.of(chunk));
+        DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(chunkContent, part.chunks);
```
jeqo, line 54: Main change

core/src/test/java/io/aiven/kafka/tieredstorage/transform/FetchChunkEnumerationTest.java
```diff
@@ -118,15 +121,19 @@ void shouldReturnRangeFromMultipleChunks() throws StorageBackendException {
         final int from = 15;
         final int to = 34;
         final FetchChunkEnumeration fetchChunk =
-            new FetchChunkEnumeration(chunkManager, SEGMENT_KEY, manifest, BytesRange.of(from, to));
-        when(chunkManager.getChunk(SEGMENT_KEY, manifest, 1))
+            new FetchChunkEnumeration(chunkManager, SEGMENT_KEY, manifest, BytesRange.of(from, to), 1);
+        final var first = new FetchPart(chunkIndex, chunkIndex.chunks().get(0), 1);
```
AnatolyPopov, line 125: `part0`?

AnatolyPopov (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/FetchPart.java
```diff
@@ -0,0 +1,140 @@
...
+
+        // set part ranges and chunk
+        this.range = BytesRange.of(firstChunk.range().from, lastChunk.range().to);
+        this.chunks = chunkIndex.chunks().subList(firstChunkId, lastChunkId + 1);
+    }
+
+    /**
+     * @see FetchPart#next()
+     */
+    private FetchPart(final ChunkIndex chunkIndex,
```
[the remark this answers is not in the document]
jeqo: It's used by the `next` method. Isn't it a valid usage?

jeqo:
Thanks @AnatolyPopov ! Adding suggestions in the last fix-up commit, have a look!

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/FetchPart.java
```diff
@@ -0,0 +1,140 @@
...
+        Objects.requireNonNull(chunk, "Chunk cannot be null");
+        if (partSize <= 0) {
+            throw new IllegalArgumentException("Part size must be higher than zero.");
+        }
+        this.partSize = partSize;
+
+        // set first and last chunk included in part, and final chunk (last in segment)
+        this.finalChunkId = chunkIndex.chunks().size() - 1;
+
+        this.firstChunkId = Math.min((chunk.id / partSize) * partSize, finalChunkId);
```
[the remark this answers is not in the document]
jeqo: Good catch. Adding a validation to fail when this is the case. Check last commit.

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/FetchManager.java
```diff
@@ -23,9 +23,9 @@
 import io.aiven.kafka.tieredstorage.storage.ObjectKey;
 import io.aiven.kafka.tieredstorage.storage.StorageBackendException;
 
-public interface ChunkManager {
-
-    InputStream getChunk(final ObjectKey objectKey,
-                         final SegmentManifest manifest,
-                         final int chunkId) throws StorageBackendException, IOException;
+public interface FetchManager {
+    InputStream partContent(final ObjectKey objectKey,
```
HenryCaiHaiying, line 27: Maybe the method should be called `fetchPartContent`?

jeqo: Sure, good idea.

ivanyu:
Do I understand that we not only fetch by these bigger parts, but also cache by them?

jeqo:
@ivanyu Correct, interaction with storage (and cache) is now based on _fetch parts_ instead of individual chunks.

jeqo:
Closing in favor of #394 #429

PR #411, opened 2023-10-02
feat: pre-fetch parts 

### LINKED ISSUE #404: Allow fetch caching to complete even if input stream closes
We have experience issues when fetch.max.wait.ms is too small for tiered data to be served. The default config value is 500ms and it may not be enough for S3 to return a result and make it available to the broker.
The problem gets worse when consumer retries and the data has to be fetched again from S3.

A potential solution, would be to allow the caching mechanism to continue even when the input stream gets closed, so the next request will find it on the cache, and will have better changes to succeed.

jeqo:
Depends on #403 

Fixes #404 

Applies approach proposed by @AnatolyPopov on #394 to prepare cache asynchronously when fetching. This will avoid failing on fetch request timeout when remote-tier fetch hasn't complete yet. 

It only pre-fetches 2 parts, the current and next one. Considering that parts are going to be either equal or greater than max fetch bytes requested per partition, this should be enough for following fetch requests to always hit the cache. 

Another side-effect is that latency may improve as remote-tier fetches happen asynchronously at the beginning of the fetch instead of on fetch read.

See last commit.


jeqo:
Closing in favor of #394 #429

PR #415, opened 2023-10-06
feat: group indexes to save remote tier calls

jeqo:
Includes usage of tagged fields to group indexes on the same map; Serialization/Deserialization; and Provider (caching).

Follows the same approaches as custom metadata for serialization/deserialization, and segment provider and chunk management for caching.


jeqo:
Closing in favor of #416

PR #416, opened 2023-10-09
feat: concat indexes in the same object

jeqo:
By having indexes together, the number of objects stored in remote tier got reduced from 4 or 5 to 1. Range fetch used to get specific indexes.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -20,9 +20,13 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.SequenceInputStream;
 import java.nio.file.Files;
 import java.security.KeyPair;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -46,6 +50,8 @@ import io.aiven.kafka.tieredstorage.chunkmanager.ChunkManagerFactory;
 import io.aiven.kafka.tieredstorage.config.RemoteStorageManagerConfig;
 import io.aiven.kafka.tieredstorage.manifest.SegmentEncryptionMetadata;
 import io.aiven.kafka.tieredstorage.manifest.SegmentEncryptionMetadataV1;
+import io.aiven.kafka.tieredstorage.manifest.SegmentIndexesV1;
+import io.aiven.kafka.tieredstorage.manifest.SegmentIndexesV1Builder;
 import io.aiven.kafka.tieredstorage.manifest.SegmentManifest;
 import io.aiven.kafka.tieredstorage.manifest.SegmentManifestProvider;
 import io.aiven.kafka.tieredstorage.manifest.SegmentManifestV1;
@@ -88,11 +94,6 @@ import org.slf4j.LoggerFactory;
 import static io.aiven.kafka.tieredstorage.config.RemoteStorageManagerConfig.METRICS_NUM_SAMPLES_CONFIG;
 import static io.aiven.kafka.tieredstorage.config.RemoteStorageManagerConfig.METRICS_RECORDING_LEVEL_CONFIG;
 import static io.aiven.kafka.tieredstorage.config.RemoteStorageManagerConfig.METRICS_SAMPLE_WINDOW_MS_CONFIG;
-import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.LEADER_EPOCH;
-import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.OFFSET;
-import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.PRODUCER_SNAPSHOT;
-import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.TIMESTAMP;
-import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType.TRANSACTION;
 
 public class RemoteStorageManager implements org.apache.kafka.server.log.remote.storage.RemoteStorageManager {
     private static final Logger log = LoggerFactory.getLogger(RemoteStorageManager.class);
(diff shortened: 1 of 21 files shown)


core/src/test/java/io/aiven/kafka/tieredstorage/manifest/SegmentManifestV1SerdeTest.java
ivanyu, line 1: It would be great to have the test for serialization/deserialization without the transaction index. Theoretically, it should serialize as `null` and deserialize just OK, but it would be great to check.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1Builder.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.aiven.kafka.tieredstorage.storage.BytesRange;
+
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType;
+
+public class SegmentIndexesV1Builder {
+    final Map<IndexType, SegmentIndexV1> indexes = new HashMap<>(IndexType.values().length);
+    int currentPosition = 0;
+
+    public SegmentIndexesV1Builder add(final IndexType indexType, final BytesRange range) {
```
ivanyu, line 30: It seems from the range we only need the size. Maybe we should pass it instead? This will also eliminate the problem of the need to check for overlapping ranges, extra tests for this, etc.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -301,35 +355,30 @@ private void uploadSegmentLog(final RemoteLogSegmentMetadata remoteLogSegmentMet
...
-                                 final SegmentEncryptionMetadata encryptionMetadata,
-                                 final SegmentCustomMetadataBuilder customMetadataBuilder)
-        throws StorageBackendException, IOException {
-        TransformChunkEnumeration transformEnum = new BaseTransformChunkEnumeration(index);
+    private InputStream transformIndex(final IndexType indexType,
+                                       final InputStream index,
+                                       final int size,
+                                       final SegmentEncryptionMetadata encryptionMetadata,
+                                       final SegmentIndexesV1Builder segmentIndexBuilder) {
+        TransformChunkEnumeration transformEnum = new BaseTransformChunkEnumeration(index, Integer.MAX_VALUE / 2);
```
ivanyu, line 363: Will this work?
```suggestion
        TransformChunkEnumeration transformEnum = new BaseTransformChunkEnumeration(index, size);
```

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
ivanyu, line 1: [Nit] Would be great to not have these extra reformattings. They seem valid (?) but should probably be a separate PR.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -415,6 +469,8 @@ public InputStream fetchIndex(final RemoteLogSegmentMetadata remoteLogSegmentMet
             }
             final DetransformFinisher detransformFinisher = new DetransformFinisher(detransformEnum);
             return detransformFinisher.toInputStream();
+        } catch (final RemoteResourceNotFoundException e) {
+            throw e;
```
ivanyu, line 473: This is redundant

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -401,8 +450,13 @@ public InputStream fetchIndex(final RemoteLogSegmentMetadata remoteLogSegmentMet
...
             final var segmentManifest = fetchSegmentManifest(remoteLogSegmentMetadata);
 
-            final var key = objectKey(remoteLogSegmentMetadata, ObjectKeyFactory.Suffix.fromIndexType(indexType));
-            final var in = fetcher.fetch(key);
+            final var key = objectKey(remoteLogSegmentMetadata, ObjectKeyFactory.Suffix.INDEXES);
+            final var segmentIndex = segmentManifest.segmentIndexes().segmentIndex(indexType);
+            if (segmentIndex == null) {
+                throw new RemoteResourceNotFoundException("Index " + indexType + " not found on " + key);
+            }
+            final var range = BytesRange.ofFromPositionAndSize(segmentIndex.position(), segmentIndex.size());
```
ivanyu, line 458: We can probably add `segmentIndex.range()` or something of this sort, but your call :)

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: not really. As the throw is within a try-catch, is captured by the last `catch Exception` and turned into a RSMException -- breaking the exception handling on the broker side.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: Actually, it does! Thanks for the suggestion.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
jeqo: Sure, good idea. Will add it.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexV1.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package io.aiven.kafka.tieredstorage.manifest;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SegmentIndexV1 implements SegmentIndex {
```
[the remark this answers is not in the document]
jeqo: Added #417

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
AnatolyPopov: I believe if we want to have it like this we should do it once and for all and hide the commit from the history, like we did in some other places.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1Builder.java
```diff
@@ -0,0 +1,90 @@
...
+        switch (indexType) {
+            case OFFSET:
+                if (!indexes.isEmpty() || currentPosition != 0) {
+                    throw new IllegalStateException("Indexes not empty, offset index must be first");
+                }
```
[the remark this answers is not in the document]
jeqo: Yeah, wasn't sure order if order would be relevant here. Indeed, it may not matter. Will remove it.

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
jeqo: Sure, leaving this out. We can follow @AnatolyPopov approach on another PR.

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
ivanyu: Yes, you're right!

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1Builder.java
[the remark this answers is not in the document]
ivanyu: I think the whole switch could be replaced with
```java
        indexes.put(indexType, new SegmentIndexV1(currentPosition, size));
        currentPosition += size;
```

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1Builder.java
```diff
@@ -0,0 +1,75 @@
...
+package io.aiven.kafka.tieredstorage.manifest;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType;
+
+public class SegmentIndexesV1Builder {
+    final Map<IndexType, SegmentIndexV1> indexes = new HashMap<>(IndexType.values().length);
+    int currentPosition = 0;
```
ivanyu, line 26: These guys should be private, I guess

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1Builder.java
```diff
@@ -0,0 +1,75 @@
...
+                indexes.put(IndexType.LEADER_EPOCH, new SegmentIndexV1(currentPosition, size));
+                currentPosition += size;
+                break;
+            case TRANSACTION:
+                indexes.put(IndexType.TRANSACTION, new SegmentIndexV1(currentPosition, size));
+                currentPosition += size;
+                break;
+            default:
+                throw new IllegalStateException("Unknown index type");
+        }
```
ivanyu, line 55: I think the whole switch could be replaced with
```suggestion
        indexes.put(indexType, new SegmentIndexV1(currentPosition, size));
        currentPosition += size;
```

jeqo: of course, 🤦🏽

ivanyu (changes requested):
LGTM, @jeqo feel free to squash

ivanyu (approved):

PR #420, opened 2023-10-12
fix: capture GCS metrics when bucket contains hyphens

jeqo:
GCS bucket names can contain hyphens, underscores, dots, etc: https://cloud.google.com/storage/docs/buckets

Fixing the regex to match URLs, and a couple of minor related refactoring. See commits for details.

### DIFF
--- a/storage/gcs/src/main/java/io/aiven/kafka/tieredstorage/storage/gcs/MetricCollector.java
+++ b/storage/gcs/src/main/java/io/aiven/kafka/tieredstorage/storage/gcs/MetricCollector.java
@@ -45,24 +45,24 @@ class MetricCollector {
      *
      * <p>That is, {@literal "/storage/v1/b/<bucket>/o/<object>"}.
      */
-    private static final Pattern OBJECT_METADATA_PATH_PATTERN =
-        Pattern.compile("^/storage/v1/b/\\w+/o/\\w+/?$");
+    static final Pattern OBJECT_METADATA_PATH_PATTERN =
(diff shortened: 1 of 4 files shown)


jeqo:
Weirdly enough, seems that the GCS Fake Server doesn't allow these characters on the bucket name, so I wasn't able to change the test; but tried it manually on our demo where it wasn't working.

storage/gcs/src/main/java/io/aiven/kafka/tieredstorage/storage/gcs/MetricCollector.java
```diff
@@ -46,23 +46,23 @@ class MetricCollector {
      * <p>That is, {@literal "/storage/v1/b/<bucket>/o/<object>"}.
      */
     private static final Pattern OBJECT_METADATA_PATH_PATTERN =
-        Pattern.compile("^/storage/v1/b/\\w+/o/\\w+/?$");
+        Pattern.compile("^/storage/v1/b/(.*)/o/(.*)/?$");
```
ivanyu, line 49: I guess this would be more correct
```suggestion
        Pattern.compile("^/storage/v1/b/([^/]+)/o/([^/]+)/?$");
```

ivanyu (changes requested):
Thanks for catching!

jeqo:
Thanks @ivanyu, suggestion makes sense. Adding a test case for the regex at least so we can validate the expressions.

ivanyu (approved):

PR #422, opened 2023-10-17
feat: validate index size when transforming

jeqo:
Index size can potentially be zero or too large for transformation to work.

### DIFF
--- a/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
+++ b/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
@@ -20,6 +20,7 @@ import javax.crypto.Cipher;
 import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


ivanyu (approved):

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
```diff
@@ -465,6 +469,79 @@ void testRequiresCompression(final CompressionType compressionType, final boolea
         assertThat(requires).isEqualTo(expectedResult);
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void testTransformingIndexes(final boolean encryption) {
```
AnatolyPopov, line 474: I'd say these tests should not be in this class but somewhere in `/test` directory since these are plain unit tests.

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
AnatolyPopov, line 1: The if statements based on parameterized tests arguments is no the best approach and raise a question about overall feasibility of the tests being parameterized but since we had it here earlier let's address it later altogether.

AnatolyPopov (approved):
In general LGTM but should be refactored later.

PR #425, opened 2023-10-20
Use disk-based cache in E2E tests

ivanyu:
The disk-based is more complex and also more realistic to be used in production.


jeqo (approved):
LGTM

PR #428, opened 2023-10-30
refactor: config prefetching based on max bytes size

jeqo:


jeqo:
Closing in favor of #429

PR #429, opened 2023-10-30
Prefetching by the number of bytes

AnatolyPopov:


core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -33,9 +33,9 @@ public class ChunkCacheConfig extends AbstractConfig {
             + "where \"-1\" represents infinite retention";
     private static final long DEFAULT_CACHE_RETENTION_MS = 600_000;
 
-    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetching.size";
+    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetching.bytes";
```
jeqo, line 36: We are using `.size` for most bytes configurations. Also, final size is not specifically this value but whatever chunks can make up to this size, right? Finally, `ing` suffix doesn't add much so we may as well drop it:
```suggestion
    private static final String CACHE_PREFETCH_MAX_SIZE_CONFIG = "prefetch.max.size";
```

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
```diff
@@ -108,6 +109,17 @@ public Chunk findChunkForOriginalOffset(final int offset) {
         );
     }
 
+    @Override
+    public List<Chunk> listChunksForRange(final BytesRange bytesRange) {
```
jeqo, line 113: nit: wdyt about:
```suggestion
    public List<Chunk> chunksForRange(final BytesRange bytesRange) {
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
AnatolyPopov: It's not max size but rather min, since we are prefetching the whole chunk if requested range ends or starts in the middle of the chunk.

AnatolyPopov: Oh actually you are right, it's max size

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
```diff
@@ -243,7 +243,7 @@ void prefetchingWholeSegment() throws Exception {
             chunkCache.configure(Map.of(
                 "retention.ms", "-1",
                 "size", "-1",
-                "prefetching.size", SEGMENT_MANIFEST.chunkIndex().chunks().size()
+                "prefetching.bytes", 20
```
ivanyu: It would be nice to bind this `20` to some named constant

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
```diff
@@ -213,7 +213,7 @@ void prefetchingNextChunk() throws Exception {
             chunkCache.configure(Map.of(
                 "retention.ms", "-1",
                 "size", "-1",
-                "prefetching.size", "1"
+                "prefetching.bytes", "10"
```
ivanyu: It would be nice to bind this `10` to some named constant

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
AnatolyPopov: Fixed

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
```diff
@@ -108,6 +109,17 @@ public Chunk findChunkForOriginalOffset(final int offset) {
...
+        Chunk current;
+        final var result = new ArrayList<Chunk>();
+        for (int i = bytesRange.from; i < bytesRange.to && i < originalFileSize; i += current.originalSize) {
```
ivanyu, line 116:
```suggestion
        for (int i = bytesRange.from; i <= bytesRange.to && i < originalFileSize; i += current.originalSize) {
```
I guess? Since `to` is inclusive.

ivanyu (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
```diff
@@ -108,6 +109,17 @@ public Chunk findChunkForOriginalOffset(final int offset) {
         );
     }
 
+    @Override
+    public List<Chunk> chunksForRange(final BytesRange bytesRange) {
+        Chunk current;
+        final var result = new ArrayList<Chunk>();
+        for (int i = bytesRange.from; i < bytesRange.to && i < originalFileSize; i += current.originalSize) {
```
ivanyu, line 116: Probably needs protection from `current == null`?

ivanyu (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
AnatolyPopov: It's not, e.g. chukSize=10, chunk1 start position will be 11, we want to prefetch single chunk => prefetch.max.size=10, so range will be 11-21. If we change to `<=` the result will contain 2 chunks. However we can make it inclusive so the range should be 11-20 then in this example.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
AnatolyPopov: Is not `i < originalFileSize` enough? We return `null` from `findChunkForOriginalOffset` only for `offset >= originalFileSize`

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
ivanyu: Yeah, it's good to have uniformity. You probably want to use `BytesRange.ofFromPositionAndSize` instead of `BytesRange.of`.

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
ivanyu: Yes, you're right

ivanyu, line 116: Let's make it explicit: do we calculate the prefetch range on the original or transformed size?

core/src/main/java/io/aiven/kafka/tieredstorage/manifest/index/AbstractChunkIndex.java
AnatolyPopov: Fixed

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
AnatolyPopov: Fixed

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
AnatolyPopov: Fixed

ivanyu:
LGTM. If it's :+1: for @jeqo , let's squash and merge

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -33,9 +33,9 @@ public class ChunkCacheConfig extends AbstractConfig {
             + "where \"-1\" represents infinite retention";
     private static final long DEFAULT_CACHE_RETENTION_MS = 600_000;
 
-    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetching.size";
+    private static final String CACHE_PREFETCHING_SIZE_CONFIG = "prefetch.max.size";
```
jeqo, line 36: nit: could we align the constant name?
```suggestion
    private static final String CACHE_PREFETCH_MAX_SIZE_CONFIG = "prefetch.max.size";
```

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
```diff
@@ -33,9 +33,9 @@ public class ChunkCacheConfig extends AbstractConfig {
...
     private static final String CACHE_PREFETCHING_SIZE_DOC =
```
jeqo, line 37: Same here

jeqo (commented):
LGTM, just a minor comment on the constant name

core/src/main/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheConfig.java
AnatolyPopov: Fixed

jeqo (approved):

ivanyu (approved):

PR #430, opened 2023-10-30
 feat: add metrics to azure blob storage 

jeqo:
Piggyback on the Http pipeline layer of Azure Http client to instrument metrics -- similar to how Tracing instrumentation is implemented.

Tried to go on the NettyHttpClient reactor metrics (following: https://projectreactor.io/docs/netty/release/reference/index.html#_metrics_4) but metrics are around data sent/received (which may include all calls) and no path to distinguish, only host:port.


storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        @Override
+        public HttpResponse processSync(final HttpPipelineCallContext context, final HttpPipelineNextSyncPolicy next) {
+            processMetrics(context);
+            return next.processSync();
+        }
+
+        void processMetrics(final HttpPipelineCallContext context) {
+            switch (context.getHttpRequest().getHttpMethod().name()) {
```
ivanyu, line 94: All three should have a path like `/<container>/<blob>` or (`/<account>/<container>/<blob>` on Azurite). There's difference in verbs and also in query.

There is no price difference between `PutBlob`, `PutBlock`, and `PutBlockList`. All three have the same verb (`PUT`) and the same path (/account/container/blob), but differ in query params only. We probably don't need to distinct them for now.

I think it's probably not needed to prepare for some hypothetical future adding of more operation which need distinction by the path and query, but it makes sense to leave a comment about this, maybe in `AzureBlobStorage.java` to be more visible.

ivanyu (changes requested):

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
```diff
@@ -0,0 +1,140 @@
...
+
+        void processMetrics(final HttpPipelineCallContext context) {
+            final var path = context.getHttpRequest().getUrl().getPath();
+            if (pathPattern.matcher(path).matches()) {
+                switch (context.getHttpRequest().getHttpMethod()) {
+                    case GET:
+                        getObjectRequests.record();
+                        break;
+                    case PUT:
+                        final var q = context.getHttpRequest().getUrl().getQuery();
```
ivanyu, line 117: Just in case, let's guard against `null` query.

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
```diff
@@ -0,0 +1,140 @@
...
+        return Pattern.compile(exp);
+    }
+
+    MetricsPolicy policy() {
+        return new MetricsPolicy(metrics, pathPattern());
+    }
+
+    static class MetricsPolicy implements HttpPipelinePolicy {
+
+        private static final Pattern UPLOAD_QUERY_PATTERN = Pattern.compile("^comp=(?<comp>\\w+)(&)?.*");
```
ivanyu, line 71: It's maybe a bit strong assumption that it's in the beginning of the query line. Let's maybe use
```suggestion
        private static final Pattern UPLOAD_QUERY_PATTERN = Pattern.compile("comp=(?<comp>[^&]+)");
```
together with `matcher.find()` instead of `matcher.matches()`?

ivanyu (changes requested):

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
```diff
@@ -0,0 +1,140 @@
...
+
+        metrics = new org.apache.kafka.common.metrics.Metrics(
+            new MetricConfig(), List.of(reporter), Time.SYSTEM,
+            new KafkaMetricsContext("aiven.kafka.server.tieredstorage.azure")
+        );
+    }
+
+    Pattern pathPattern() {
+        final var maybeContainer = config.containerName() != null ? config.containerName() + "/" : "";
+        final var exp = "^/" + config.accountName() + "/" + maybeContainer + "([^/])";
```
ivanyu, line 61: In the normal Azure, the account is a part of the host name and doesn't appear in the path.

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
```diff
@@ -0,0 +1,140 @@
...
+        final JmxReporter reporter = new JmxReporter();
+
+        metrics = new org.apache.kafka.common.metrics.Metrics(
+            new MetricConfig(), List.of(reporter), Time.SYSTEM,
+            new KafkaMetricsContext("aiven.kafka.server.tieredstorage.azure")
+        );
+    }
+
+    Pattern pathPattern() {
+        final var maybeContainer = config.containerName() != null ? config.containerName() + "/" : "";
```
ivanyu, line 60: Can `containerName` be null?

ivanyu (changes requested):

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
ivanyu: More on this:

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/MetricCollector.java
jeqo: Got it! I mixed this up with the account name potentially being on the hostname. Fixed.

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorage.java
```diff
@@ -108,6 +114,9 @@ public long upload(final InputStream inputStream, final ObjectKey key) throws St
         parallelTransferOptions.setMaxSingleUploadSizeLong(blockSizeLong);
         final BlockBlobOutputStreamOptions options = new BlockBlobOutputStreamOptions()
             .setParallelTransferOptions(parallelTransferOptions);
+        // Be aware that metrics instrumentation is based on PutBlock (upload part)
+        // and PutBlockList (complete upload) used by this call.
+        // If upload changes, change metrics instrumentation accordingly.
```
ivanyu, line 119: I just realized that if we upload a file smaller than `azure.upload.block.size`, it will be a plain `PutBlob`, i.e. no blocks involved. I think we should adapt to this as well.

ivanyu (changes requested):

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorage.java
jeqo: Good catch! adding a test for this, and including putBlob metric.

storage/azure/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorageMetricsTest.java
```diff
@@ -0,0 +1,206 @@
...
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.InstanceOfAssertFactories.DOUBLE;
+
+@Testcontainers
+public class AzureBlobStorageMetricsTest {
```
ivanyu, line 55: It seems like a lot could be shared between this test class and `AzureBlobStorageTest`. Maybe let's extract some functions to some intermediate class?

storage/azure/src/main/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorage.java
ivanyu: The comment probably needs changing now :)

ivanyu (changes requested):

storage/azure/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorageMetricsTest.java
jeqo: Well thought. Adding a new utils for Azurite.

storage/azure/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/azure/AzureBlobStorageMetricsTest.java
```diff
@@ -0,0 +1,185 @@
...
+            .isGreaterThan(0.0);
+        assertThat(MBEAN_SERVER.getAttribute(objectName, "blob-get-total"))
+            .isEqualTo(2.0);
+
+        if (expectedPutBlob > 0) {
+            assertThat(MBEAN_SERVER.getAttribute(objectName, "blob-upload-rate"))
+                .asInstanceOf(DOUBLE)
+                .isGreaterThan(0.0);
+            assertThat(MBEAN_SERVER.getAttribute(objectName, "blob-upload-total"))
+                .isEqualTo(expectedPutBlob);
```
ivanyu, line 160: I guess this should be outside of the `if` (and others below, too)? We want to check it always.

ivanyu (changes requested):
Seems to be the last bit :)

ivanyu (commented):
LGTM, feel free to squash.

ivanyu (approved):

PR #438, opened 2023-11-07
chore: add jmh benchmarks module

jeqo:
Include a new module to host JMH benchmarks from TS plugin components and benchmarks for transform/detransform.

Follow a similar setup as the benchmarks on apache/kafka to simplify customizing arguments.


AnatolyPopov (approved):
LGTM

PR #439, opened 2023-11-07
chore: add maven-publish plugin

jeqo:
As the plugins may eventually be published for deployment or tooling that needs access to tiered data, we can start by adding maven coordinates to the project.

The initial proposal is to add modules under `io.aiven.kafka:kafka-ts-plugin-*` -- though open for other options.


jeqo:
Changing proposal to `io.aiven:tiered-storage-for-apache-kafka-*` to keep it aligned with repo naming and any trademark issues

ivanyu (approved):

PR #442, opened 2023-11-09
chore: re-enable deps validation

jeqo:


build.gradle
```diff
@@ -240,38 +240,46 @@ tasks.distTar {
...
-                            conflictsFound = true
+        def storages = ["s3", "gcs", "azure"]
+        def ignoreDeps = ["e2e"]
+
+        storages.forEach { s ->
+            println "Validating modules for storage: $s"
+            def dependencyVersions = [:]
+            subprojects { subproject ->
+                if (ignoreDeps.contains(subproject.name)) return
+                if (s == subproject.name || !storages.contains(subproject.name)) {
```
ivanyu, line 253: How do you thinks: would it be possible to declare this task on the storage project-level so we don't need to filter and search (of course from the single definition)?

jeqo: Good idea, there's only common and core modules to add as deps to storage backends. See last commit, thanks!

AnatolyPopov:
What about just using 
```
configurations.configureEach {
    resolutionStrategy {
        failOnVersionConflict()
    }
}
```
?

jeqo:
@AnatolyPopov if you reproduce the first commit in this PR, the following conflicts are found:

```
> Task :storage:azure:validateDeps FAILED
Conflict found for jackson-annotations: 2.13.5 vs 2.16.1
Conflict found for jackson-core: 2.13.5 vs 2.16.1
Conflict found for jackson-databind: 2.13.5 vs 2.16.1

FAILURE: Build failed with an exception.

* Where:
Build file '/mnt/aiven-home-dir/aiven/aiven-open/tiered-storage-for-apache-kafka/storage/build.gradle' line: 62

* What went wrong:
Execution failed for task ':storage:azure:validateDeps'.
> Dependency conflicts found!. Expression: conflictsFound. Values: conflictsFound = true

```

This is unfortunately not found by your suggested configuration. IIUC, as the storage modules are not dependent on core then the collision of libraries between these are not found by default tooling--hence the custom script.

AnatolyPopov:
Well, after looking into this more deeply I can say that default tooling does not find the conflict because there is no dependencies between `core` module and submodules of `storage`. However I believe we should have such a dependency for this use case specifically because the storage can not be used with `core`. The ideal scope for such a dependency would be maven's `provided` I guess but there is no such thing in Gradle since we are not using war plugin. So IMO the solution should be:
1. Use the default tooling as I proposed before
2. Create a `providedRuntime` scope similar to Grade war plugin e.g.
```
configurations {
    providedRuntime
    implementation.extendsFrom(providedRuntime)
}
```
3. Add a dependency in `:storage:core` like this
`providedRuntime(project(":core"))`

This kind of scope definition will not change the resulting distribution, e.g. `:core` dependencies will not be included into artifacts of submodules of `:storage` but it will make the tooling work properly. And I guess for connectors or any other similar plugins people are doing similar things. And same for web apps, since the container provides some libraries there.

AnatolyPopov:
@jeqo Please take a look https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/518

PR #452, opened 2023-11-16
refactor: renaming module chunkmanager to fetch 

jeqo:
Now that purpose of `chunkmanager` module is more clear, we can rename it to `fetch` so it's explicit when it's used and regroup the components that belong to this module (specifically FetchChunkEnumeration that does not really belong in the `transform` module). 

Changes:
- Renaming module to `fetch`
- Moving FetchChunkEnumeration from `transform` into `fetch` module
- Renaming cache configuration with `fetch` prefix

See commits for details.

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -49,8 +49,6 @@ make consume
 
 This scenario uses `S3Storage` with the real AWS S3 as the remote storage.
 
-Please note that the in-memory cache `io.aiven.kafka.tieredstorage.fetch.cache.InMemoryChunkCache` is used in this scenario. This is necessary to mitigate a now-present bug in Kafka with slow and excessive read from S3.
-
 For this scenario you need to have:
 1. Valid AWS S3 credentials (e.g. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_KEY`).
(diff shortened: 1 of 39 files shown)


demo/README.md
```diff
@@ -49,7 +49,7 @@ make consume
 
 This scenario uses `S3Storage` with the real AWS S3 as the remote storage.
 
-Please note that the in-memory cache `io.aiven.kafka.tieredstorage.chunkmanager.cache.InMemoryChunkCache` is used in this scenario. This is necessary to mitigate a now-present bug in Kafka with slow and excessive read from S3.
+Please note that the in-memory cache `io.aiven.kafka.tieredstorage.fetch.cache.InMemoryChunkCache` is used in this scenario. This is necessary to mitigate a now-present bug in Kafka with slow and excessive read from S3.
```
ivanyu, line 52: I guess this line is inaccurate in both sentences:
1. We switched to the disk cache in demos.
2. The second issue is not what the caches are against nowadays :) 

Let's remove it altogether?

ivanyu (approved):

AnatolyPopov (approved):

PR #453, opened 2023-11-20
Explicitly test against cache poisoning with failed futures

ivanyu:
This commit adds tests that ensure the failed futures in our caches don't stay there after their failure is found out. The removal is handled by Caffeine.

### DIFF
--- a/core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
+++ b/core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
@@ -336,5 +336,47 @@ class ChunkCacheTest {
                     .hasRootCauseInstanceOf(RuntimeException.class)
                     .hasRootCauseMessage(TEST_EXCEPTION_MESSAGE);
         }
+
+        @Test
+        void cacheIsNotPoisonedWithFailedFuturesOnFetching() throws Exception {
+            when(chunkManager.getChunk(eq(SEGMENT_OBJECT_KEY), eq(SEGMENT_MANIFEST), eq(0)))
+                .thenThrow(new StorageBackendException(TEST_EXCEPTION_MESSAGE))
+                .thenReturn(new ByteArrayInputStream(new byte[1]));
+
+            assertThatThrownBy(() -> chunkCache
+                .getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0))
+                .isInstanceOf(StorageBackendException.class)
+                .hasMessage(TEST_EXCEPTION_MESSAGE);
+
+            await().atMost(Duration.ofMillis(50))
+                .pollInterval(Duration.ofMillis(5))
+                .ignoreExceptions()
+                .until(() ->
+                    chunkCache.getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0).readAllBytes().length == 1);
+        }
+
+        @Test
+        void cacheIsNotPoisonedWithFailedFuturesOnPrefetching() throws Exception {
+            when(chunkManager.getChunk(eq(SEGMENT_OBJECT_KEY), eq(SEGMENT_MANIFEST), eq(0)))
+                .thenReturn(new ByteArrayInputStream(new byte[1]));
+            when(chunkManager.getChunk(eq(SEGMENT_OBJECT_KEY), eq(SEGMENT_MANIFEST), eq(1)))
+                .thenThrow(new StorageBackendException(TEST_EXCEPTION_MESSAGE))
+                .thenReturn(new ByteArrayInputStream(new byte[1]));
+            // To avoid returning null on prefetching when we fetch 1 directly:
+            when(chunkManager.getChunk(eq(SEGMENT_OBJECT_KEY), eq(SEGMENT_MANIFEST), eq(2)))
+                .thenReturn(new ByteArrayInputStream(new byte[1]));
+
+            chunkCache.configure(Map.of(
+                "retention.ms", "-1",
+                "size", "-1",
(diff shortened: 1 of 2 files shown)


core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
```diff
@@ -336,5 +336,47 @@ void failedReadingCachedValueWithExecutionException() throws Exception {
...
+            assertThatThrownBy(() -> chunkCache
+                .getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0))
+                .isInstanceOf(StorageBackendException.class)
+                .hasMessage(TEST_EXCEPTION_MESSAGE);
+
+            await().atMost(Duration.ofMillis(50))
+                .pollInterval(Duration.ofMillis(5))
+                .ignoreExceptions()
+                .until(() ->
+                    chunkCache.getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0).readAllBytes().length == 1);
```
AnatolyPopov, line 355: Do we really need it like this? I would expect an exception to be never thrown here. more over it's a second call to mock that should always return 1 byte as specified in thenReturn. So should it be just
`assertThat(chunkCache.getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0)).hasBinaryContent(new byte[1]);` ?

jeqo (approved):
LGTM, thanks @ivanyu for looking into these!

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
AnatolyPopov: Also instead of plain comparison in `until` it would be cleaner to user `untilAsserted` and use AssertJ assertion inside.

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
```diff
@@ -336,5 +336,47 @@ void failedReadingCachedValueWithExecutionException() throws Exception {
...
+                "retention.ms", "-1",
+                "size", "-1",
+                "prefetch.max.size", ORIGINAL_CHUNK_SIZE
+            ));
+            chunkCache.getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0);
+            await().atMost(Duration.ofMillis(500))
+                .pollInterval(Duration.ofMillis(5))
+                .ignoreExceptions()
+                .until(() ->
+                    chunkCache.getChunk(SEGMENT_OBJECT_KEY, SEGMENT_MANIFEST, 0).readAllBytes().length == 1);
```
AnatolyPopov, line 379: `untilAsserted` instead I guess.

AnatolyPopov: I do not really understand what are we testing here.
Should we verify that prefetching actually happened? Should we verify when and why we actually called cache or DefaultChunkManager to see that?

core/src/test/java/io/aiven/kafka/tieredstorage/manifest/SegmentManifestProviderTest.java
```diff
@@ -141,4 +142,24 @@ void shouldPropagateIOException(@Mock final InputStream isMock) throws StorageBa
...
+            .isInstanceOf(StorageBackendException.class)
+            .hasMessage("test");
+
+        final var chunkIndex = new FixedSizeChunkIndex(100, 1000, 110, 110);
+        final var expectedManifest = new SegmentManifestV1(chunkIndex, SEGMENT_INDEXES, false, null, null);
+
+        await().atMost(Duration.ofMillis(50))
+            .pollInterval(Duration.ofMillis(5))
+            .ignoreExceptions()
+            .until(() -> provider.get(MANIFEST_KEY).equals(expectedManifest));
```
AnatolyPopov, line 163: Same as in ChunkCache: seems that awaitility is not needed since the second call to the mock I assume will always return the expectedManifest.

AnatolyPopov (changes requested):

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
ivanyu: The removal of the failed future is asynchronous and it becomes flaky without waiting.

> Also instead of plain comparison in `until` it would be cleaner to user `untilAsserted` and use AssertJ assertion inside.

Good point, I'll make a PR.

core/src/test/java/io/aiven/kafka/tieredstorage/manifest/SegmentManifestProviderTest.java
ivanyu: > Same as in ChunkCache: seems that awaitility is not needed since the second call to the mock I assume will always return the expectedManifest.

Answered above: this assumption is incorrect. `.get`s in quick succession result in the more than one time return of the failed future. This is a result of the future scheduling.

core/src/test/java/io/aiven/kafka/tieredstorage/chunkmanager/cache/ChunkCacheTest.java
ivanyu: > I do not really understand what are we testing here.

We verify that failed futures are removed from the cache.

> Should we verify that prefetching actually happened? Should we verify when and why we actually called cache or DefaultChunkManager to see that?

I'm not sure. This all is tested in neighbor tests, do we actually need to re-check this?

PR #454, opened 2023-11-23
refactor: metric names with -metric suffix

jeqo:
Align metric group names to have `-metric` suffix


core/src/main/java/io/aiven/kafka/tieredstorage/fetch/cache/DiskBasedChunkCacheMetrics.java
```diff
@@ -28,7 +28,7 @@
 import org.apache.kafka.common.utils.Time;
 
 class DiskBasedChunkCacheMetrics {
-    private static final String METRIC_GROUP = "chunk-cache-disk";
+    private static final String METRIC_GROUP = "chunk-cache-disk-metrics";
```
AnatolyPopov, line 31: Should we make the rename first and then rename the metrics as well into `disk-chunk-cache-metrics` and etc.?

jeqo: Good idea, let's focus on #455  and applied suggestion as fixup commit already

storage/gcs/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/gcs/GcsStorageMetricsTest.java
```diff
@@ -108,7 +108,7 @@ void metricsShouldBeReported() throws StorageBackendException, IOException, JMEx
         storage.delete(key);
 
         final ObjectName gcsMetricsObjectName =
-            ObjectName.getInstance("aiven.kafka.server.tieredstorage.gcs:type=gcs-metrics");
+            ObjectName.getInstance("aiven.kafka.server.tieredstorage.gcs:type=gcs-storage-metrics");
```
ivanyu, line 111: Weeeell, considering `gcs` is already "storage"... :)

ivanyu:
The Kafka code base follows this convention, I think it's good we're doing this rename.

storage/gcs/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/gcs/GcsStorageMetricsTest.java
jeqo: s3 + storage = s4 😂

jeqo: Going for s3-client-metrics instead. Let me know how it looks now

AnatolyPopov (dismissed):
LGTM

AnatolyPopov (approved):

PR #456, opened 2023-11-23
feat: return a known exception when remote fetch is interrupted

jeqo:
This new exception aims to identify interrupted remote fetch requests caused by small fetch.max.wait.ms configurations.
In the case of fetching logs, if it happens when caching a chunk or fetching a segment manifest, it will fail with an empty stream that will lead to a retry.
If it fails when fetching an index, then it's exception will be triggered, but instead of a generic RuntimeException, a RemoteFetchTimeoutException is thrown.


jeqo:
Closing in favor of https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/462

PR #469, opened 2023-12-13
 fix: add empty indexes to builder 

jeqo:
By ignoring empty indexes--that is a valid case when segment sizes are too small; even if not recommended--the requirement to have at least 4 indexes when uploaded was not met.

This fix improves the logging to show what indexes are included therefore know which one is missing; and includes a fix to add empty indexes to the builder.

See commits for more details. It includes moving some RSM tests from IT to unit tests.

### DIFF
--- a/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
+++ b/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
@@ -501,6 +501,37 @@ class RemoteStorageManagerTest extends RsaKeyAwareTest {
             segmentIndexBuilder
         );
         assertThat(is).isNotEmpty();
+        assertThat(segmentIndexBuilder.indexes()).containsOnly(IndexType.OFFSET);
+
+
+        // adding required indexes to test builder
+        rsm.transformIndex(
+            IndexType.TIMESTAMP,
+            new ByteArrayInputStream(bytes),
+            bytes.length,
+            encryptionMetadata,
+            segmentIndexBuilder
+        );
+        rsm.transformIndex(
+            IndexType.LEADER_EPOCH,
+            new ByteArrayInputStream(bytes),
+            bytes.length,
+            encryptionMetadata,
+            segmentIndexBuilder
+        );
+        rsm.transformIndex(
+            IndexType.PRODUCER_SNAPSHOT,
+            new ByteArrayInputStream(bytes),
+            bytes.length,
+            encryptionMetadata,
+            segmentIndexBuilder
+        );
+        final var index = segmentIndexBuilder.build();
+        assertThat(index.offset().size()).isGreaterThan(0);
+        assertThat(index.timestamp().size()).isGreaterThan(0);
+        assertThat(index.leaderEpoch().size()).isGreaterThan(0);
+        assertThat(index.producerSnapshot().size()).isGreaterThan(0);
+        assertThat(index.transaction()).isNull();
     }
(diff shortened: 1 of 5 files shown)


core/src/test/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1BuilderTest.java
```diff
@@ -39,7 +39,8 @@ void shouldFailWhenBuildingWithNotEnoughIndexes() {
             .add(RemoteStorageManager.IndexType.TIMESTAMP, 1)
             .build())
             .isInstanceOf(IllegalStateException.class)
-            .hasMessage("Not enough indexes have been added. At least 4 required");
+            .hasMessage("Not enough indexes have been added; at least 4 required. "
+                + "Indexes included: [TIMESTAMP, OFFSET]");
```
AnatolyPopov, line 43: This one is flaky: the order of index types in the exception message is not guaranteed.

AnatolyPopov (changes requested):

core/src/test/java/io/aiven/kafka/tieredstorage/manifest/SegmentIndexesV1BuilderTest.java
jeqo: Good catch! Turning it into a sorted list, have a look at the last commit.

ivanyu:
A question about [refactor: further moving integration to unit tests for RSM](https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/469/commits/cf581ec3b86a8e5cad494cb9c7847aa4fd629b55): what's the purpose of this move? AFAICS, we still use the same `FileSystemStorage`, so it's still a running real RSM with real backend. Or I'm missing some other change?

ivanyu:
Apart from that question, LGTM

jeqo:
> AFAICS, we still use the same FileSystemStorage, so it's still a running real RSM with real backend. 

Oh, I was under the impression that we would keep only long-running tests on integration tests. Let's move them back, with the additional changes.

jeqo:
@ivanyu RSMTest changes are in place now, have a look. Thanks!

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
```diff
@@ -532,6 +533,7 @@ void testTransformingEmptyIndexes(final boolean encryption) {
             segmentIndexBuilder
         );
         assertThat(is).isEmpty();
+        assertThat(segmentIndexBuilder.indexes()).containsOnly(IndexType.OFFSET);
```
ivanyu, line 536: Should we maybe add `segmentIndexBuilder.build()` here and check the result, at least that the indexed size is indeed 0?

ivanyu (changes requested):

core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
jeqo: Sure, adding validations on the latest commit

ivanyu (changes requested):
LGTM! 
@AnatolyPopov do you have anything to add? If not, @jeqo feel free to squash, let's merge this

ivanyu (approved):

AnatolyPopov (approved):

PR #443, opened 2023-11-10
Refactor chunk cache

ivanyu:
Before, the chunk cache(s) was a wrapper over `DefaultChunkManager`. In some situations, this is not very convenient and also the cache has more responsibilities than it should (like prefetching). This commit refactors the cache to be a pluggable mechanism in the chunk manager.

PR #472, opened 2023-12-22
 feat: add async cache for segment indexes 

jeqo:
Indexes are currently cached on broker-side but with a synchronous cache that is not able to source value when interrupted.
This new cache will fetch index asynchronously, and retries would have chances to find the value sourced already.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -52,8 +52,11 @@ import io.aiven.kafka.tieredstorage.fetch.ChunkManager;
 import io.aiven.kafka.tieredstorage.fetch.ChunkManagerFactory;
 import io.aiven.kafka.tieredstorage.fetch.FetchChunkEnumeration;
 import io.aiven.kafka.tieredstorage.fetch.KeyNotFoundRuntimeException;
+import io.aiven.kafka.tieredstorage.fetch.index.MemorySegmentIndexesCache;
+import io.aiven.kafka.tieredstorage.fetch.index.SegmentIndexesCache;
 import io.aiven.kafka.tieredstorage.manifest.SegmentEncryptionMetadata;
 import io.aiven.kafka.tieredstorage.manifest.SegmentEncryptionMetadataV1;
+import io.aiven.kafka.tieredstorage.manifest.SegmentIndex;
 import io.aiven.kafka.tieredstorage.manifest.SegmentIndexesV1;
 import io.aiven.kafka.tieredstorage.manifest.SegmentIndexesV1Builder;
 import io.aiven.kafka.tieredstorage.manifest.SegmentManifest;
@@ -122,6 +125,7 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
     private Set<SegmentCustomMetadataField> customMetadataFields;
 
     private SegmentManifestProvider segmentManifestProvider;
+    private SegmentIndexesCache segmentIndexesCache;
 
     public RemoteStorageManager() {
         this(Time.SYSTEM);
@@ -167,6 +171,9 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
             fetcher,
             mapper,
             executor);
+
+        segmentIndexesCache = new MemorySegmentIndexesCache();
+        segmentIndexesCache.configure(config.fetchIndexesCacheConfigs());
 
         customMetadataSerde = new SegmentCustomMetadataSerde();
         customMetadataFields = config.customMetadataKeysIncluded();
@@ -513,25 +520,47 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
             if (segmentIndex == null) {
                 throw new RemoteResourceNotFoundException("Index " + indexType + " not found on " + key);
             }
-            final var in = fetcher.fetch(key, segmentIndex.range());
-
-            DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(in);
-            final Optional<SegmentEncryptionMetadata> encryptionMetadata = segmentManifest.encryption();
-            if (encryptionMetadata.isPresent()) {
-                detransformEnum = new DecryptionChunkEnumeration(
-                    detransformEnum,
-                    encryptionMetadata.get().ivSize(),
-                    encryptedChunk -> aesEncryptionProvider.decryptionCipher(encryptedChunk, encryptionMetadata.get())
-                );
-            }
-            final DetransformFinisher detransformFinisher = new DetransformFinisher(detransformEnum);
-            return detransformFinisher.toInputStream();
+            return segmentIndexesCache.get(
+                key,
+                indexType,
+                () -> fetchIndexBytes(key, segmentIndex, segmentManifest)
+            );
         } catch (final RemoteResourceNotFoundException e) {
             throw e;
         } catch (final KeyNotFoundException e) {
             throw new RemoteResourceNotFoundException(e);
         } catch (final Exception e) {
             throw new RemoteStorageException(e);
+        }
+    }
+
+    private byte[] fetchIndexBytes(
+        final ObjectKey key,
+        final SegmentIndex segmentIndex,
+        final SegmentManifest segmentManifest
+    ) {
+        final InputStream in;
+        try {
+            in = fetcher.fetch(key, segmentIndex.range());
+        } catch (final StorageBackendException e) {
+            throw new RuntimeException("Error fetching index from remote storage", e);
+        }
+
+        DetransformChunkEnumeration detransformEnum = new BaseDetransformChunkEnumeration(in);
+        final Optional<SegmentEncryptionMetadata> encryptionMetadata = segmentManifest.encryption();
+        if (encryptionMetadata.isPresent()) {
+            detransformEnum = new DecryptionChunkEnumeration(
+                detransformEnum,
+                encryptionMetadata.get().ivSize(),
+                encryptedChunk -> aesEncryptionProvider.decryptionCipher(encryptedChunk,
+                    encryptionMetadata.get())
+            );
+        }
+        final var detransformFinisher = new DetransformFinisher(detransformEnum);
+        try (final var is = detransformFinisher.toInputStream()) {
+            return is.readAllBytes();
+        } catch (final IOException e) {
+            throw new RuntimeException("Error reading de-transformed index bytes", e);
         }
     }
 
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2024 Aiven Oy
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -513,19 +520,35 @@ public InputStream fetchIndex(final RemoteLogSegmentMetadata remoteLogSegmentMet
...
+                            encryptedChunk -> aesEncryptionProvider.decryptionCipher(encryptedChunk,
+                                encryptionMetadata.get())
+                        );
+                    }
+                    final var detransformFinisher = new DetransformFinisher(detransformEnum);
+                    try (final var is = detransformFinisher.toInputStream()) {
+                        return is.readAllBytes();
+                    } catch (final IOException e) {
+                        throw new RuntimeException("Error reading de-transformed index bytes", e);
+                    }
```
ivanyu, line 549: [Nit] I'd probably move this to a separate static method

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCache.java
```diff
@@ -0,0 +1,130 @@
...
+import com.github.benmanes.caffeine.cache.Scheduler;
+import com.github.benmanes.caffeine.cache.Weigher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MemorySegmentIndexesCache implements SegmentIndexesCache {
+    private static final Logger log = LoggerFactory.getLogger(MemorySegmentIndexesCache.class);
+
+    private static final long GET_TIMEOUT_SEC = 10;
+    private static final long DEFAULT_MAX_SIZE_BYTES = 100 * 1024 * 1024;
```
ivanyu, line 51: This looks a bit too much (even considering the cache has TTL). Considering these entries are subsequently cached on the broker side when successfully returned, maybe we should have something smaller, e.g. 10 MiB?

core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalLong;
+
+import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.ConfigDef;
+
+import static org.apache.kafka.common.config.ConfigDef.NO_DEFAULT_VALUE;
+
+public class CacheConfig extends AbstractConfig {
```
ivanyu, line 29: I guess, there's some plan to make this generic for all caches?

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCache.java
```diff
@@ -0,0 +1,130 @@
...
+    protected AsyncCache<SegmentIndexKey, byte[]> cache;
+
+    public RemovalListener<SegmentIndexKey, byte[]> removalListener() {
+        return (key, content, cause) -> log.debug("Deleted cached value for key {} from cache."
+            + " The reason of the deletion is {}", key, cause);
+    }
+
+    public Weigher<SegmentIndexKey, byte[]> weigher() {
+        return (key, value) -> value.length;
+    }
```
ivanyu, line 66: Do these two need to be higher-order functions?
```suggestion
    private static void removalListener(final SegmentIndexKey key, final byte[] value, RemovalCause cause) {
        log.debug("Deleted cached value for key {} from cache. The reason of the deletion is {}", key, cause);
    }

    private static int weigher(final SegmentIndexKey key, byte[] value) {
        return value.length;
    }
```

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/index/SegmentIndexKey.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public String toString() {
+        return "SegmentIndexKey{"
+            + "indexesKey=" + indexesKey
+            + ", indexType=" + indexType
+            + '}';
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
ivanyu, line 43: We probably need another test like those `*EqualsTest`

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+            final InputStream cachedOffsetIndex = cache.get(
+                key,
+                IndexType.OFFSET,
+                offsetIndexSupplier
+            );
+            assertThat(cachedOffsetIndex).hasBinaryContent(OFFSET_INDEX);
+            verifyNoMoreInteractions(offsetIndexSupplier);
+            assertThat(cache.cache.asMap()).hasSize(1);
+
+            Thread.sleep(100);
```
ivanyu, line 178: It's worth adding a comment that this sleep is needed to spread away in time the creation (and eviction) of the cache entries, or something like this

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+            final ObjectKey key = objectKeyFactory.key(REMOTE_LOG_SEGMENT_METADATA, ObjectKeyFactory.Suffix.INDEXES);
+            assertThatThrownBy(() -> cache
+                .get(key, IndexType.OFFSET, offsetIndexSupplier))
+                .isInstanceOf(StorageBackendException.class)
+                .hasMessage(TEST_EXCEPTION_MESSAGE);
+            assertThatThrownBy(() -> cache
+                .get(key, IndexType.TIMESTAMP, timeIndexSupplier))
+                .isInstanceOf(IOException.class)
+                .hasMessage(TEST_EXCEPTION_MESSAGE);
+        }
```
ivanyu, line 289: [Nit] This probably asks for splitting into two tests

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCache.java
jeqo: Agree, reducing to 10mib

core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
jeqo: Yeap, will do a follow-up PR to do this and include refactoring ideas from https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/443

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
jeqo: Sure. Turning this into a parametrized test.

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
AnatolyPopov: Instead of just waiting, should we wait for removal listener to be called or for size becoming zero?

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+                    eq(RemovalCause.EXPIRED));
+        }
+
+        @Test
+        void sizeBasedEviction() throws IOException, StorageBackendException {
+            cache.configure(Map.of(
+                "size", "18",
+                "retention.ms", "-1"
+            ));
+            assertThat(cache.cache.asMap()).hasSize(0);
```
AnatolyPopov, line 218: `isEmpty()`? and in other places. Also perhaps `isNotEmpty()` for `.size()).isGreaterThanOrEqualTo(1)`?

core/src/test/java/io/aiven/kafka/tieredstorage/config/CacheConfigTest.java
```diff
@@ -0,0 +1,134 @@
...
+class CacheConfigTest {
+
+    @Test
+    void cacheUnboundedSize() {
+        final CacheConfig config = new CacheConfig(
+            new ConfigDef(),
+            Map.of("size", "-1")
+        );
+
+        assertThat(config.cacheSize()).isEmpty();
```
AnatolyPopov, line 39: `isNotPresent()` instead would improve readability a lot

core/src/main/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCache.java
```diff
@@ -0,0 +1,131 @@
...
+
+            throw new RuntimeException(e);
+        } catch (final InterruptedException | TimeoutException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void configure(final Map<String, ?> configs) {
+        final var config = new CacheConfig(new ConfigDef(), configs, DEFAULT_MAX_SIZE_BYTES);
```
AnatolyPopov, line 128: Do we really need to instantiate `ConfigDef` here? Maybe move it inside `CacheConfig`?

core/src/main/java/io/aiven/kafka/tieredstorage/config/RemoteStorageManagerConfig.java
```diff
@@ -41,6 +41,7 @@
 
 public class RemoteStorageManagerConfig extends AbstractConfig {
     private static final String STORAGE_PREFIX = "storage.";
+    public static final String FETCH_INDEXES_CACHE_PREFIX = "fetch.indexes.cache.";
```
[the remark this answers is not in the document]
jeqo: Good catch. This was following a similar approach as cache factory config, but I agree we could instead create a method in the config itself.

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
jeqo: @AnatolyPopov, falling back to using thread sleep (same as in ChunkCacheTest) as adding a stronger condition (e.g. waiting for cache to be empty) takes unknown amount of time github actions--unless there's a better approach here.

ivanyu (dismissed):
LGTM

core/src/test/java/io/aiven/kafka/tieredstorage/fetch/index/MemorySegmentIndexesCacheTest.java
AnatolyPopov: I see you fixed it here but what about other places?

jeqo: Found it. Should be fixed on the latest commit

AnatolyPopov (approved):

PR #480, opened 2024-01-09
feat: extend BytesRange to support empty ranges

jeqo:
Given that indexes could be empty, inclusive ranges should be able to represent the case where size is zero. 
To do this, `to=-1` is proposed to represent ranges that are empty.

Historically, ranges were designed only for segments where empty files were not possible. As recently empty indexes became a possibility when small segment files are created, then the initial design of BytesRange became an issue as it's designed as inclusive-inclusive.


ivanyu:
One of the reasons why inclusive-inclusive range bounds are not popular in the industry :D

ivanyu (approved):

giuseppelillo (approved):

PR #490, opened 2024-02-02
feat: extend object deleter to delete multiple keys at once when supported by back-end

### LINKED ISSUE #319: Use DeleteObjects to delete whole segment faster and with fewer requests
https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html

Don't forget to add the new request to https://github.com/aiven/tiered-storage-for-apache-kafka/blob/052e425d753535fd3ae880ffd6d7e839392e9415/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/MetricCollector.java#L41

jeqo:
Some Storage back-ends have the ability to delete multiple objects at once. To allow them to use this optimization, the ObjectDeleter interface is extended to allow passing multiple object keys at once.

Fix #319

PS. Was considering GCS initially, but given that FakeGcs server doesn't seem to properly handle the `delete(List of BlobId)` -- at least metrics where not processed -- and implementation seems to be syntactic sugar; then I decided to postpone it and only focused on S3.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -26,6 +26,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.KeyPair;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


AnatolyPopov (approved):
LGTM

PR #507, opened 2024-03-11
Wrap more S3 errors in StorageBackendException

ivanyu:
There's no reason in passing raw HTTP/IO and other similar exceptions upstream.


jeqo (approved):

AnatolyPopov (approved):
LGTM

PR #509, opened 2024-03-12
Remove HTTPS from Azure tests

ivanyu:
Apparently, with all the credentials used in the tests, plain HTTP is fine. Removing HTTPS will simplify some future tests.


jeqo (approved):
LGTM

PR #514, opened 2024-03-14
chore(build): build docker image via gradle

jeqo:
To reduce friction on developers wanting to build the plugin, this PR adds docker plugin to gradle and define it as a dependency for e2e tests.


OneCricketeer:
Why not use Jib Gradle plugin?

PR #516, opened 2024-03-16
Close file InputStreams when copying segments

### LINKED ISSUE #513: Disk space not released
# What can we help you with?
I found that after running for a while, the disk occupancy is not actually released, I am using raft cluster mode and I don't know how to troubleshoot it
image
Why is this happening? I found that after I restarted the node again, the disk space was freed up

# Where would you expect to find this information?

ivanyu:
Closes #513

It's easier to review `RemoteStorageManager.java` if you hide white spaces in Github or your IDE.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
@@ -223,25 +223,29 @@ public class RemoteStorageManager implements org.apache.kafka.server.log.remote.
         final long startedMs = time.milliseconds();
 
         try {
-            TransformChunkEnumeration transformEnum = new BaseTransformChunkEnumeration(
-                Files.newInputStream(logSegmentData.logSegment()), chunkSize);
             SegmentEncryptionMetadataV1 encryptionMetadata = null;
             final boolean requiresCompression = requiresCompression(logSegmentData);
-            if (requiresCompression) {
-                transformEnum = new CompressionChunkEnumeration(transformEnum);
-            }
-            if (encryptionEnabled) {
-                final DataKeyAndAAD dataKeyAndAAD = aesEncryptionProvider.createDataKeyAndAAD();
-                transformEnum = new EncryptionChunkEnumeration(
-                    transformEnum,
-                    () -> aesEncryptionProvider.encryptionCipher(dataKeyAndAAD));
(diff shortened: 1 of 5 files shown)


funky-eyes (approved):
LGTM

jeqo (approved):
LGTM, nicely solved.

Thanks @funky-eyes for reporting, and @ivanyu for the fix!

funky-eyes:
@jeqo When can this PR be merged?

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
```diff
@@ -223,23 +223,26 @@ public Optional<CustomMetadata> copyLogSegmentData(final RemoteLogSegmentMetadat
...
+                    encryptionMetadata = new SegmentEncryptionMetadataV1(dataKeyAndAAD.dataKey, dataKeyAndAAD.aad);
+                }
+                transformFinisher = new TransformFinisher(transformEnum, remoteLogSegmentMetadata.segmentSizeInBytes());
+                uploadSegmentLog(remoteLogSegmentMetadata, transformFinisher, customMetadataBuilder);
             }
-            final TransformFinisher transformFinisher =
-                new TransformFinisher(transformEnum, remoteLogSegmentMetadata.segmentSizeInBytes());
-            uploadSegmentLog(remoteLogSegmentMetadata, transformFinisher, customMetadataBuilder);
 
             final ChunkIndex chunkIndex = transformFinisher.chunkIndex();
```
AnatolyPopov, line 247: Should not it be so that the lifecycle of transformFinisher should be same as logSementInputStream? Saving chunkIndex into filed in rather a side effect, it is safe to get it here but the main purpose of the finisher is based on the underlying stream that is already closed afaiu. So I'd rather would not expose finisher from the try related to underlying stream or maybe would even try to make it auto closeable(not sure if it's possible) and close simultaneously with the underlying stream.

AnatolyPopov: In general LGTM but not sure if we need a nested try, imo it could be just single one.

AnatolyPopov (approved):

core/src/main/java/io/aiven/kafka/tieredstorage/RemoteStorageManager.java
AnatolyPopov: And please squash

ivanyu: Done

jeqo (approved):

PR #518, opened 2024-03-18
Validating runtime dependencies for conflicts

AnatolyPopov:


build.gradle
```diff
@@ -74,6 +74,11 @@ subprojects {
         withJavadocJar()
         withSourcesJar()
     }
+    configurations.runtimeClasspath {
```
jeqo, line 77: Could we add some comments here to clarify how is affecting the deps validation? e.g. usage of platform feature, etc.

build.gradle
```diff
@@ -123,29 +128,32 @@ subprojects {
 
         azureSdkVersion = "1.2.21"
 
-        testcontainersVersion = "1.19.4"
+        testcontainersVersion = "1.19.7"
```
jeqo, line 131: was this needed as part of this PR? if it's maybe a separate commit will help.

build.gradle
```diff
@@ -123,29 +128,32 @@ subprojects {
...
 
         testcontainersFakeGcsServerVersion = "0.2.0"
     }
 
     dependencies {
         compileOnly "org.apache.kafka:kafka-clients:$kafkaVersion"
-        compileOnly "org.apache.kafka:kafka-storage-api:$kafkaVersion"
+        compileOnly("org.apache.kafka:kafka-storage-api:$kafkaVersion")
```
jeqo, line 138: wondering if this is a style change only or it has some implications in the context of this PR?

build.gradle
```diff
@@ -123,29 +128,32 @@ subprojects {
...
 
+        implementation enforcedPlatform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")
```
jeqo, line 140: could we add some comments on what this "enforcePlatform" does and why we needed it here?

e.g. as it's only added for jacksone, does it mean that we would need to know in advance which dependencies may conflict, so we can flag it as we are doing here?

storage/plugins/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+subprojects {
+    configurations {
+        providedRuntime
+        implementation.extendsFrom(providedRuntime)
+    }
```
jeqo, line 21: maybe some comments here as well

jeqo (changes requested):
@AnatolyPopov thanks for looking into this.

I like that it's more idiomatic now using gradle APIs to enforce the dependencies; but this also makes it harder to reason about if don't know the gradle internals. Hence I'd ask for some additional comments to clarify how to maintain this approach in the future.

About the "plugins" submodule, I'd say we may be overusing the term in other places (the project itself is sometimes referred as plugin). Could we consider `backend` instead as it's also known in the codebase?

What do you think about re-structuring the commits in a way that it's clear to see how enabling this new validation strategy flags conflicting dependencies? similar to https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/442/commits.

I think we are also missing cleaning up the previous validation scripts. Could we include it on this PR as well?

PR #519, opened 2024-03-19
Include AWS STS in S3 storage distro

### LINKED ISSUE #511: Add AWS STS to allow using AWS creds from service account
# What is currently missing?

Using

`rsm.config.aws.credentials.provider.class=software.amazon.awssdk.auth.credentials.WebIdentityTokenFileCredentialsProvider` 

 deals with reading the environment variables that are injected to EKS pod and assuming the role with STS. 
 
 Kafka logs show:

`WARN To use web identity tokens, the 'sts' service module must be on the class path. (software.amazon.awssdk.auth.credentials.internal.WebIdentityCredentialsUtils)` and authentication failed.

# How could this be improved?

Internally, `WebIdentityTokenFileCredentialsProvider` will call AWS STS so you will need to add it to the dependencies list.

# Is this a feature you would work on yourself?

* [ ] I plan to open a pull request for this feature

ivanyu:
STS is a very popular credentials provider. It's good to have it in the S3 storage distribution.

Resolves #511

### DIFF
--- a/storage/s3/build.gradle
+++ b/storage/s3/build.gradle
@@ -23,15 +23,12 @@ ext {
 dependencies {
     implementation project(":storage:core")
 
-    // Take out jackson libraries out to avoid collision with core jackson dependencies
-    implementation ("software.amazon.awssdk:s3:$awsSdkVersion") {
-        exclude group: "com.fasterxml.jackson.core"
-        exclude group: "com.fasterxml.jackson.dataformat"
-        exclude group: "org.slf4j"
+    def excludeFromAWSDeps = { ModuleDependency dep ->
+        dep.exclude group: "org.slf4j"
     }
-    // replacing with version used by core
-    implementation "com.fasterxml.jackson.core:jackson-databind:$jacksonVersion"
-    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:$jacksonVersion"
+    implementation ("software.amazon.awssdk:s3:$awsSdkVersion") {excludeFromAWSDeps(it)}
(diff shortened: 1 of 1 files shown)


jeqo (dismissed):
LGTM

storage/s3/build.gradle
```diff
@@ -24,11 +24,14 @@ dependencies {
     implementation project(":storage:core")
 
     // Take out jackson libraries out to avoid collision with core jackson dependencies
-    implementation ("software.amazon.awssdk:s3:$awsSdkVersion") {
-        exclude group: "com.fasterxml.jackson.core"
-        exclude group: "com.fasterxml.jackson.dataformat"
-        exclude group: "org.slf4j"
+    def excludeFromAWSDeps = { ModuleDependency dep ->
+        dep.exclude group: "com.fasterxml.jackson.core"
+        dep.exclude group: "com.fasterxml.jackson.dataformat"
```
AnatolyPopov, line 29: Instead of using this excludes everywhere should we just add
` implementation enforcedPlatform("com.fasterxml.jackson:jackson-bom:$jacksonVersion")`
in subprojects in parent pom?

ivanyu: In fact, Jackson doesn't need exclusion, because the AWS libraries has it shadowed.
And SLF4J needs exactly exclusion, because we don't include in on the core as well (expect it to be in Kafka).
Please check the fixup

AnatolyPopov: LGTM, however we could replace SLF4J exclusion globally for `runtimeOnly` on configuration level to avoid copy-paste. Will make a PR later. Please squash.

ivanyu: Done

AnatolyPopov (approved):
LGTM

Issue #520, opened 2024-03-21
How to limit the speed of tiered storage


funky-eyes:
# What can we help you with?
When reading and writing in real time, whenever the action of uploading s3 is triggered, it will lead to a very large disk io consumption, cpu instantly rises, resulting in the impact of the production of the business operation, I tried to adjust remote.log.manager.thread.pool.size=cpu*2 but the effect is still not very ideal
![image](https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/assets/19943636/be457b88-cc1e-4e7a-9000-c6baff554317)
![image](https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/assets/19943636/c6732d28-e86f-49bc-92ac-a2ead7fab6dd)

# Where would you expect to find this information?

funky-eyes:
I want to do speed limit in S3MultiPartOutputStream#write, I wonder if the community can agree to this solution, I can mention pr to support it

ivanyu:
It seems you will need [KIP-956](https://cwiki.apache.org/confluence/display/KAFKA/KIP-956+Tiered+Storage+Quotas), which hasn't been accepted yet.

funky-eyes:
> It seems you will need [KIP-956](https://cwiki.apache.org/confluence/display/KAFKA/KIP-956+Tiered+Storage+Quotas), which hasn't been accepted yet.

@ivanyu Yes, I need this feature, but I found that I can actually achieve rate limiting in the implementation code of copysegment. So why don't we implement this feature ourselves? Just like how the community provides compression for copysegment.

ivanyu:
Yeah, I see your point. We're looking into your PRs and other ways to implement this

funky-eyes:
@ivanyu 
https://github.com/apache/kafka/pull/15625 
I've noticed this standard RLMQuotaManager, but it seems that we still need to adapt and use it because it appears to be just a standard definition.

jeqo:
@funky-eyes I have been looking into how quotas are implemented and it's not going to help us that much on the upload path as the granularity is at the segment level. It will not help on _how fast_ a segment is actually read and uploaded. For the fetch part it will be more effective I think.

[...]

PR #521, opened 2024-03-21
support s3 upload rate limiting

funky-eyes:
As a result of triggering the upload of s3, a large number of files are read without any limitations resulting in the disk io directly hit the full, producer to send a message is very prone to timeout
Resolves: #520
Under my testing, the cpu trend is very smooth, the disk disk iops are reduced from a peak of 3000 to about 1500, and the disk io is at 20%, so if you tune some more flow limiting values, you can basically do tiered storage within another safe system load value
image
image


funky-eyes:
I think the implementation of #548 is more elegant, so I plan to close this PR.

PR #524, opened 2024-03-28
Add experimental SOCKS5 support for S3

ivanyu:
Technically, I found a way to have proxy-side host name resolution, but it requires rewriting quite a bit of the AWS client machinery. Leaving it as is for now.


jeqo (approved):
Thanks @ivanyu. LGTM overall.

nit: I wonder if there's any consideration/reminder we should be comment about the usage of reflection. As this is an experimental feature it may be ok as is though.

ivanyu:
@jeqo I isolated all the reflection operations in a separate package + added a comment on this. Even when this becomes non-experimental, most likely the reflection magic stays: the AWS SDK isn't just configurable enough to do what we want as we did with the other clouds.

PR #531, opened 2024-04-04
feat: remove orphan objects if copy fails at any point

jeqo:
Given that there are at least 3 objects uploaded per segment, and it could fail at any stage, the plugin should guard that the chances for orphan objects is minimal.

Be aware, the first commit includes a set of refactoring around RSM upload operations to cleanup and simplify the testing of the 2nd (main) commit. See commit messages for more details.

### DIFF
--- a/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
+++ b/core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
@@ -57,7 +57,6 @@ import org.apache.kafka.server.log.remote.storage.RemoteStorageManager.IndexType
 import io.aiven.kafka.tieredstorage.fetch.KeyNotFoundRuntimeException;
 import io.aiven.kafka.tieredstorage.fetch.cache.DiskChunkCache;
 import io.aiven.kafka.tieredstorage.fetch.cache.MemoryChunkCache;
-import io.aiven.kafka.tieredstorage.manifest.SegmentEncryptionMetadataV1;
 import io.aiven.kafka.tieredstorage.manifest.SegmentIndexesV1Builder;
 import io.aiven.kafka.tieredstorage.manifest.index.ChunkIndex;
 import io.aiven.kafka.tieredstorage.manifest.serde.EncryptionSerdeModule;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/manifest/SegmentManifestV1.java
```diff
@@ -41,19 +41,19 @@ public SegmentManifestV1(
...
     ) {
-        this(chunkIndex, segmentIndexes, compression, encryption, null);
+        this(chunkIndex, segmentIndexes, compression, Optional.ofNullable(encryption), null);
     }
 
     public SegmentManifestV1(final ChunkIndex chunkIndex,
                              final SegmentIndexesV1 segmentIndexes,
                              final boolean compression,
-                             final SegmentEncryptionMetadataV1 encryption,
+                             final Optional<SegmentEncryptionMetadataV1> encryption,
```
ivanyu, line 50: Not that I'm a dogmatist, but even Brian Goetz [advises](https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555) against using `Optional` as args :smiley:

jeqo: That's fair. I was on the fence about this one. For private methods (where I started introducing it) looks like a better input type--even though as Brian mentions it's not intended as a "Maybe" type (not serializable, etc)--but it leaked to this public method. 
I'm ok falling back to the previous type. Do you think is still valid type for private methods or should we undo those too?

jeqo: I have undo changes for internal methods as well. In the case of SegmentManifestV1, I think it's a good candidate to introduce a Builder for it.
I have also spot a couple of input parameters using Optional in our code base, will do a quick refactor later to clean those up.

jeqo: Follow ups:
- https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/532 to clean existing optional input types
- https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/533 to have a manifest builder

jeqo:
Closing this as it has got too old. Will create a fresh PR, soon

PR #532, opened 2024-04-08
refactor: Remove Optional input parameters

jeqo:
Cleanup code accepting Optional types, see https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/pull/531#discussion_r1553816416

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
@@ -19,7 +19,6 @@ package io.aiven.kafka.tieredstorage.config;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
-import java.util.OptionalLong;
 
 import org.apache.kafka.common.config.AbstractConfig;
 import org.apache.kafka.common.config.ConfigDef;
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
```diff
@@ -60,11 +56,11 @@ private static ConfigDef addCacheConfigs(final OptionalLong maybeDefaultSize) {
...
 
     public CacheConfig(final Map<String, ?> props) {
-        super(addCacheConfigs(OptionalLong.empty()), props);
+        super(configDef(NO_DEFAULT_VALUE), props);
     }
 
-    public CacheConfig(final Map<String, ?> props, final long defaultSize) {
-        super(addCacheConfigs(OptionalLong.of(defaultSize)), props);
+    public CacheConfig(final Map<String, ?> props, final Long defaultSize) {
+        super(configDef(Objects.requireNonNull(defaultSize, "defaultSize cannot be null")), props);
```
ivanyu, line 63: If it can't be `null`, should we keep `long` as the type here?

ivanyu (changes requested):

core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
jeqo: Good catch. Added a Builder to conditionally pass this value when needed. PTAL.

core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
```diff
@@ -82,4 +77,22 @@ public Optional<Duration> cacheRetention() {
         }
         return Optional.of(Duration.ofMillis(rawValue));
     }
+
+    public static class Builder {
+        final Map<String, ?> props;
+        Object maybeDefaultSize = NO_DEFAULT_VALUE;
```
ivanyu, line 83:
```suggestion
        private final Map<String, ?> props;
        private Object maybeDefaultSize = NO_DEFAULT_VALUE;
```
I think

ivanyu (changes requested):
LGTM, one small nit

ivanyu (approved):

PR #537, opened 2024-04-09
feat: add upload from file to storage backend 

jeqo:
Backend SDKs may have optimized ways to upload directly from files, instead of loading to memory as part of the plugin.


storage/core/src/main/java/io/aiven/kafka/tieredstorage/storage/ObjectUploader.java
```diff
@@ -16,13 +16,31 @@
...
+import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 public interface ObjectUploader {
     /**
      * @param inputStream content to upload. Not closed as part of the upload.
-     * @param key         path to an object within a storage backend.
+     * @param key         destintion path to an object within a storage backend.
```
Mwea, line 27:
```suggestion
     * @param key         destination path to an object within a storage backend.
```

storage/core/src/main/java/io/aiven/kafka/tieredstorage/storage/ObjectUploader.java
```diff
@@ -16,13 +16,31 @@
...
     long upload(InputStream inputStream, ObjectKey key) throws StorageBackendException;
+
+    /**
+     * Enable backend to use optimized uploading implementations based on source files
+     *
+     * @param path source path to the object to upload
+     * @param size size of the object to upload
+     * @param key  destination path to an object within a storage backend
+     */
+    default long upload(Path path, int size, ObjectKey key) throws StorageBackendException {
```
Mwea, line 39: Where is this `upload` call used ?

jeqo: It will be in a following PR (updated the PR description to make this clear), it touches RemoteStorageManager logic that is being refactored in other PRs (e.g. #531)

PR #549, opened 2024-05-22
fix(s3): reducing array copying on S3 output stream

jeqo:
Use only ByteRange slices to pass bytes to S3 client operations and remove array copying.

Doing some benchmarks on the current implementation, the Arrays.copyOfRange dominates the memory allocation:

image

 By switching to the ByteBuffer approach, this copying is removed:

image

### DIFF
--- a/storage/s3/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageTest.java
+++ b/storage/s3/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageTest.java
@@ -24,6 +24,7 @@ import io.aiven.kafka.tieredstorage.storage.StorageBackend;
 import io.aiven.kafka.tieredstorage.storage.TestObjectKey;
 import io.aiven.kafka.tieredstorage.storage.TestUtils;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -50,8 +51,8 @@ public class S3StorageTest extends BaseStorageTest {
 
     @BeforeAll
     static void setUpClass() {
-        final var clientBuilder = S3Client.builder();
-        clientBuilder.region(Region.of(LOCALSTACK.getRegion()))
+        s3Client = S3Client.builder()
+            .region(Region.of(LOCALSTACK.getRegion()))
             .endpointOverride(LOCALSTACK.getEndpointOverride(LocalStackContainer.Service.S3))
             .credentialsProvider(
                 StaticCredentialsProvider.create(
@@ -62,13 +63,19 @@ public class S3StorageTest extends BaseStorageTest {
                 )
             )
             .build();
-        s3Client = clientBuilder.build();
     }
 
     @BeforeEach
     void setUp(final TestInfo testInfo) {
         bucketName = TestUtils.testNameToBucketName(testInfo);
         s3Client.createBucket(CreateBucketRequest.builder().bucket(bucketName).build());
+    }
+
+    @AfterAll
+    static void closeAll() {
+        if (s3Client != null) {
+            s3Client.close();
+        }
     }
 
     @Override
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
@@ -169,8 +169,11 @@ public class S3MultiPartOutputStream extends OutputStream {
 
     private void flushBuffer(final int offset,
                              final int actualPartSize) {
-        final ByteArrayInputStream in = new ByteArrayInputStream(partBuffer.array(), offset, actualPartSize);
-        uploadPart(in, actualPartSize);
+        try (final ByteArrayInputStream in = new ByteArrayInputStream(partBuffer.array(), offset, actualPartSize)) {
+            uploadPart(in, actualPartSize);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
         partBuffer.clear();
     }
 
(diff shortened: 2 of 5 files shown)


funky-eyes (approved):
LGTM

funky-eyes:
However, in Java, the close method of ByteArrayInputStream has no effect. The methods of this class can be called after the stream has been closed without generating an IOException. This is because the data of ByteArrayInputStream is stored in memory, unlike file streams or network streams that require actual resource cleanup, so there may not be an out-of-memory problem

jeqo:
@funky-eyes good catch! I'm adding more changes on how bytes are to the request using byte buffers only instead of array copying. PTAL

funky-eyes (approved):
LGTM

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -93,16 +94,23 @@ public void write(final byte[] b, final int off, final int len) throws IOExcepti
...
-                processedBytes += transferred;
-                source.position(source.position() + transferred);
+            final ByteBuffer inputBuffer = ByteBuffer.wrap(b, off, len);
+            while (inputBuffer.hasRemaining()) {
+                // copy batch to part buffer
+                final int inputLimit = inputBuffer.limit();
+                final int toCopy = Math.min(partBuffer.remaining(), inputBuffer.remaining());
+                final int positionAfterCopying = inputBuffer.position() + toCopy;
+                inputBuffer.limit(positionAfterCopying);
+                partBuffer.put(inputBuffer.slice());
```
jeqo, line 104: Main change: remove array copying, and reuse input buffer

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -167,11 +178,14 @@ private void abortUpload() {
...
                              final int actualPartSize) {
-        final ByteArrayInputStream in = new ByteArrayInputStream(partBuffer.array(), offset, actualPartSize);
-        uploadPart(in, actualPartSize);
-        partBuffer.clear();
+        try (final InputStream in = new ByteBufferInputStream(buffer)) {
+            processedBytes += actualPartSize;
+            uploadPart(in, actualPartSize);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
```
jeqo, line 188: Also, provide an inputstream directly from buffer stream instead of creating a new byte array.

jeqo:
For reference, similar improvements have been implemented on kafka core: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/actions/runs/10206672059/job/28240015688?pr=576

storage/s3/src/integration-test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageTest.java
```diff
@@ -50,8 +50,8 @@ public class S3StorageTest extends BaseStorageTest {
 
     @BeforeAll
     static void setUpClass() {
-        final var clientBuilder = S3Client.builder();
-        clientBuilder.region(Region.of(LOCALSTACK.getRegion()))
+        s3Client = S3Client.builder()
```
biggusdonzus, line 53: nit: it probably doesn't matter for a test, but i think that `s3Client` needs to be closed in a `@AfterAll` method to cleanup eventual connections/resources.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -54,12 +54,13 @@ public void configure(final Map<String, ?> configs) {
...
     public long upload(final InputStream inputStream, final ObjectKey key) throws StorageBackendException {
-        try (final var out = s3OutputStream(key)) {
+        final var out = s3OutputStream(key);
+        try (out) {
             inputStream.transferTo(out);
-            return out.processedBytes();
         } catch (final IOException e) {
             throw new StorageBackendException("Failed to upload " + key, e);
         }
+        return out.processedBytes();
```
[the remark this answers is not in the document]
jeqo: Added a comment to clarify a bit, but the main reason this is now outside the resource try-catch is because the `processedBytes` in the output stream is counted now as part of the `flushBuffer` and not the `write` method anymore. 
This is more accurate as it counts bytes processed _after upload_ compared to the previous behavior where it was count when writing.
Before, the processed bytes would be known before closing. Now as the last `flushBuffer` may occur when closing the stream, it means that the result is not known before closing, thus the change.

biggusdonzus (approved):
looks good :+1:

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
```diff
@@ -186,19 +189,29 @@ void writesOneByte() throws Exception {
...
+        final List<RequestBody> requestBodies = new ArrayList<>();
         when(mockedS3.uploadPart(any(UploadPartRequest.class), any(RequestBody.class)))
             .thenAnswer(invocation -> {
-                final UploadPartRequest up = invocation.getArgument(0);
-                return newUploadPartResponse("SOME_ETAG#" + up.partNumber());
+                final UploadPartRequest upload = invocation.getArgument(0);
+                final RequestBody originalBody = invocation.getArgument(1);
+                //emulate the behavior of S3 client otherwise we will get a wrong array in the memory
+                try (final InputStream inputStream = originalBody.contentStreamProvider().newStream()) {
+                    final RequestBody requestBody = RequestBody.fromBytes(inputStream.readAllBytes());
```
jomach, line 201: Why not RequestBody.fromInputStream() ?

jomach: Well I just saw that it is only on tests so I would say not relevant :)

PR #558, opened 2024-06-06
docs: Increase cache.retention.ms configuration example

funky-eyes:
Incorporate the `cache.retention.ms` configurat ion into the documentation to highlight its purpose. Otherwise, it's easy to overlook this configuration, leading to suboptimal parameter tuning.

Resolves: Increase cache.retention.ms configuration example

In the context of an offline data analysis business scenario, where computation times for business loads are lengthy, improper configuration of cache retention time can lead to frequent interactions with S3 for data block retrieval, consequently escalating the costs associated with S3. However, this configuration is not currently documented and is only discovered through reading the source code. Hence, there's a desire to include it in the documentation, enabling users to optimize it according to their specific scenarios.


showuon (approved):
LGTM!

jeqo (approved):
Thanks @funky-eyes !

PR #559, opened 2024-06-06
[datadog] Allocate direct buffers for multipart upload

dopuskh3:
Allow to allocate multipart upload buffers as direct buffer rather than on the heap.

## Why?

We try to set a pretty large multipart upload part size on cluster to optimize throughput and reduce S3 requests.

At the same time, we try to keep kafka JVM heap size contained on most kafka installation in order to leave as much memory as possible for the page cache. As a matter of example, we will use 4GB heap size on machines with 64GB available memory.

The consequence of using pretty large multipart upload size on contained JVM heap size is that we can pretty easily run out of heap size if we suddenly have to upload many segments to tiered storage.

The strategy we propose is to allocate multipart buffer in direct memory so that we can more easily configure direct buffer budget.

Resolves: #xxxxx
Why this way


funky-eyes:
I think this feature is good.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -44,12 +44,15 @@ public class S3Storage implements StorageBackend {
     private String bucketName;
     private int partSize;
 
+    private boolean multiartDirectBuffers;
```
rsharan-dd, line 47:
```suggestion
    private boolean multipartDirectBuffers;
```

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -44,12 +44,15 @@ public class S3Storage implements StorageBackend {
...
+
     @Override
     public void configure(final Map<String, ?> configs) {
         final S3StorageConfig config = new S3StorageConfig(configs);
         this.s3Client = S3ClientBuilder.build(config);
         this.bucketName = config.bucketName();
         this.partSize = config.uploadPartSize();
+        this.multiartDirectBuffers = config.multipartDirectBuffers();
```
rsharan-dd, line 55:
```suggestion
        this.multipartDirectBuffers = config.multipartDirectBuffers();
```

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -63,7 +66,7 @@ public long upload(final InputStream inputStream, final ObjectKey key) throws St
     }
 
     S3MultiPartOutputStream s3OutputStream(final ObjectKey key) {
-        return new S3MultiPartOutputStream(bucketName, key, partSize, s3Client);
+        return new S3MultiPartOutputStream(bucketName, key, partSize, s3Client, multiartDirectBuffers);
```
rsharan-dd, line 69:
```suggestion
        return new S3MultiPartOutputStream(bucketName, key, partSize, s3Client, multipartDirectBuffers);
```

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -58,10 +58,16 @@ public class S3StorageConfig extends AbstractConfig {
...
+
+    public static final String S3_MULTIPART_UPLOAD_DIRECT_BUFFERS_CONFIG = "s3.multipart.upload.direct.buffers";
+    public static final String S3_MULTIPART_UPLOAD_DIRECT_BUFFERS_DOC =
+        "Allocate multipart upload buffers as direct buffers (off-heap)";
+
     static final int S3_MULTIPART_UPLOAD_PART_SIZE_MIN = 5 * 1024 * 1024; // 5MiB
     static final int S3_MULTIPART_UPLOAD_PART_SIZE_MAX = Integer.MAX_VALUE;
     static final int S3_MULTIPART_UPLOAD_PART_SIZE_DEFAULT = S3_MULTIPART_UPLOAD_PART_SIZE_MIN;
 
+
```
rsharan-dd, line 70: minor nit: unnecessary whitespace

rsharan-dd (changes requested):

dopuskh3:
@jeqo fixed tests, PTAL

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStreamTest.java
jeqo, line 1: Could these tests be parametrized to have coverage for the direct allocation?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -44,12 +44,15 @@ public class S3Storage implements StorageBackend {
     private String bucketName;
     private int partSize;
 
+    private boolean multipartDirectBuffers;
```
jeqo, line 47: nit:
```suggestion
    private boolean bufferAllocationDirect;
```

jeqo (commented):
thanks @dopuskh3! I agree this is a valuable flag for deployments where JVM memory may not be enough for the part uploads. 
I added few minor naming suggestions and some testing coverage; but overall looks good.

PR #571, opened 2024-07-17
fix: exclude unnecessary modules in Caffeine

showuon:
In the released core zip, we included 
1. caffeine-3.1.8.jar
2. checker-qual-3.37.0.jar
3. error_prone_annotations-2.21.1.jar
From this discussion in Caffeine project: https://github.com/ben-manes/caffeine/issues/300#issuecomment-507292996 . The author of Caffeine project:
 
> checker-qual And errorprone are annotations which Java doesn’t require at runtime unless you reflectively load them when inspecting the methods. Unfortunately other languages will not classload without them, making them required. You can exclude safely in Java.

Since we only use JAVA and it doesn't cause any errors while compiling after excluding them, I think we should remove them from the distribution zip.

### DIFF
--- a/storage/gcs/build.gradle
+++ b/storage/gcs/build.gradle
@@ -20,8 +20,12 @@ dependencies {
     implementation project(":storage:core")
 
     implementation ("com.google.cloud:google-cloud-storage:$gcpSdkVersion") {
-        exclude group: "com.google.errorprone"
-        exclude group: "org.checkerframework"
+        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
+        exclude group: 'org.checkerframework', module: 'checker-qual'
+        exclude group: 'com.google.code.findbugs', module: 'jsr305'
+        exclude group: 'com.google.j2objc', module: 'j2objc-annotations'
+        exclude group: 'org.codehaus.mojo', module: 'animal-sniffer-annotations'
+        exclude group: 'com.google.guava', module: 'listenablefuture'
     }
 
     implementation project(":commons")
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -41,7 +41,10 @@ dependencies {
 
     implementation "com.github.luben:zstd-jni:$zstdVersion"
 
-    implementation "com.github.ben-manes.caffeine:caffeine:$caffeineVersion"
+    implementation("com.github.ben-manes.caffeine:caffeine:$caffeineVersion") {
+        exclude group: 'org.checkerframework', module: 'checker-qual'
+        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
+    }
 
     implementation "commons-io:commons-io:$apacheCommonsIOVersion"
     implementation project(':commons')


jeqo:
Thanks @showuon! Good finding, and agree it should help to make the distribution smaller and including only needed deps.
On the issue referenced, it's mentioned that Scala may not be happy if we remove these. The kafka broker root classes are still in Scala, however the plugin is loaded via Java classloading. I'm don't have much experience here to say if this is an issue or not. Could you elaborate on why this is not an issue? I guess as e2e tests are passing fine then it should be ok.
Also, the issue also references guava that is used by the google cloud sdk. I guess we could also consider removing those?

showuon:
> I'm don't have much experience here to say if this is an issue or not. Could you elaborate on why this is not an issue? I guess as e2e tests are passing fine then it should be ok.

TBH, I don't know much detail about it. But I agree with you that as long as the e2e tests passed, it means we don't need to ship these modules.

> Also, the issue also references guava that is used by the google cloud sdk. I guess we could also consider removing those?

I agree. I just used the latest main to build gcs, it didn't include the package. I think it's because we've excluded them from gcs here: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/commit/63dcb12668cca76f0c754ca4c98f716395a5e4e6.

jeqo:
@showuon, yes that PR removed some, but in the [issue comments](https://github.com/ben-manes/caffeine/issues/300#issuecomment-507291069), it references a few others: 

```
        <dependency>
            <groupId>com.google.guava</groupId>
            <artifactId>guava</artifactId>
            <version>28.0-jre</version>
            <exclusions>
                <exclusion>
                    <groupId>com.google.code.findbugs</groupId>
                    <artifactId>jsr305</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>org.checkerframework</groupId>
                    <artifactId>checker-qual</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>com.google.errorprone</groupId>
                    <artifactId>error_prone_annotations</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>com.google.j2objc</groupId>
                    <artifactId>j2objc-annotations</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>org.codehaus.mojo</groupId>
                    <artifactId>animal-sniffer-annotations</artifactId>
                </exclusion>
                <exclusion>
                    <groupId>com.google.guava</groupId>
                    <artifactId>listenablefuture</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```
I think it should be safe to exclude them as well?

storage/gcs/build.gradle
```diff
@@ -20,8 +20,12 @@ dependencies {
...
 
     implementation ("com.google.cloud:google-cloud-storage:$gcpSdkVersion") {
-        exclude group: "com.google.errorprone"
-        exclude group: "org.checkerframework"
+        exclude group: 'com.google.errorprone', module: 'error_prone_annotations'
+        exclude group: 'org.checkerframework', module: 'checker-qual'
+        exclude group: 'com.google.code.findbugs', module: 'jsr305'
+        exclude group: 'com.google.j2objc', module: 'j2objc-annotations'
+        exclude group: 'org.codehaus.mojo', module: 'animal-sniffer-annotations'
+        exclude group: 'com.google.guava', module: 'listenablefuture'
```
showuon, line 28: Include module name to be more specific.

showuon:
Good point @jeqo ! PR updated. Let's see if e2e tests can pass.

showuon:
All tests passed!

jeqo (approved):
LGTM, thanks @showuon !

PR #590, opened 2024-09-02
feat: remove orphan objects if copy fails at any point

jeqo:
Given that there are at least 3 objects uploaded per segment, and it could fail at any stage, the plugin should guard that the chances for orphan objects is minimal.


biggusdonzus (commented):
Looks good :+1:

biggusdonzus (approved):

PR #591, opened 2024-09-11
refactor: reuse cache config on chunk cache

jeqo:
Extend cache config with chunk cache configs (default and disk-based config) to reuse the configuration definitions.


.github/workflows/main_push_and_pull_request_workflow.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ 'main' ]
   pull_request:
-    branches: [ 'main' ]
```
jeqo, line 7: Adding this in a separate commit to allow other chained PRs to build fully instead of waiting for main to be the base branch.

Issue #584, opened 2024-08-30
many error logs output when fetching data from AWS S3


showuon:
# What happened?

When fetch data from AWS S3, I saw many error messages with interrupt exception. Although the data is fetched successfully and data is correct, I think we should try to fix the issue. If it's expected, maybe we should catch the exception and do some custom handling?

configs:
```
[2024-08-30 15:04:52,250] INFO RemoteStorageManagerConfig values: 
	chunk.size = 4194304
	compression.enabled = false
	compression.heuristic.enabled = false
	custom.metadata.fields.include = []
	encryption.enabled = false
	key.prefix = 
	key.prefix.mask = false
	metrics.num.samples = 2
	metrics.recording.level = INFO
	metrics.sample.window.ms = 30000
	segment.manifest.cache.retention.ms = 3600000
	segment.manifest.cache.size = 1000
	storage.backend.class = class io.aiven.kafka.tieredstorage.storage.s3.S3Storage
 (org.apache.kafka.common.config.AbstractConfig)
[2024-08-30 15:04:52,253] INFO S3StorageConfig values: 
	aws.access.key.id = [hidden]
	aws.certificate.check.enabled = true
	aws.checksum.check.enabled = false
	aws.credentials.provider.class = null
	aws.secret.access.key = [hidden]
	s3.api.call.attempt.timeout = null
	s3.api.call.timeout = null
	s3.bucket.name = [hidden]
	s3.endpoint.url = null
	s3.multipart.upload.part.size = 5242880
	s3.path.style.access.enabled = null
	s3.region = ap-southeast-2
 (org.apache.kafka.common.config.AbstractConfig)
```

errors logs: https://gist.github.com/showuon/8ff9e18b062c7392c6e2b189f045dc3f

# What did you expect to happen?

No error logs.

# What else do we need to know?

N/A

jeqo:
Thanks @showuon! This may be related with KIP-1018. We have this ticket to track the fix and document it: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/issues/483 
Could you give a try to the new configuration on the broker side? Setting `remote.fetch.max.wait.ms` to a few seconds should be enough to avoid the interruptions.

dopuskh3:
@showuon I ran into the same issue when `remote.fetch.max.wait.ms` was too.

[...]

PR #592, opened 2024-09-11
feat: additional configs for caches

jeqo:
Currently the ForkJoinPools used by the fetch caches (chunk, indexes, and manifest) are defined by default. Also, the fetching is timeout by default to 10secs.
This PR aims to introduce thread-pool size and timeout configurations to the cache config to each fetch cache, to allow tuning.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
@@ -31,6 +31,12 @@ public class CacheConfig extends AbstractConfig {
     private static final String CACHE_RETENTION_CONFIG = "retention.ms";
     private static final String CACHE_RETENTION_DOC = "Cache retention time ms, "
         + "where \"-1\" represents infinite retention";
+    private static final String CACHE_FETCH_THREAD_POOL_SIZE_CONFIG = "fetch.thread.pool.size";
+    private static final String CACHE_FETCH_THREAD_POOL_SIZE_DOC = "Size for the thread pool used to "
+        + "schedule asynchronous fetching tasks, default to number of processors.";
+    private static final String CACHE_FETCH_TIMEOUT_MS_CONFIG = "fetch.timeout.ms";
+    private static final String CACHE_FETCH_TIMEOUT_MS_DOC = "When getting an object from the fetch, "
+        + "how long to wait before timing out. Defaults to 10 sec.";
 
     static final long CACHE_RETENTION_MS_DEFAULT = 600_000;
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
```diff
@@ -55,6 +61,22 @@ private static ConfigDef configDef(
             ConfigDef.Importance.MEDIUM,
             CACHE_RETENTION_DOC
         );
+        configDef.define(
+            CACHE_FETCH_THREAD_POOL_SIZE_CONFIG,
+            ConfigDef.Type.INT,
+            0,
+            ConfigDef.Range.between(0, 1024),
```
AnatolyPopov, line 68: 1024 looks quite big, should it be smaller?

jeqo: Yeah, didn't want it be to be too constrained 😅 ok to shrink it, do you have a better value in mind?

jeqo: Found that the limit on the ForkJoinPool is 32767
```
static final int MAX_CAP      = 0x7fff;        // max #workers - 1
```
so, maybe 1024 (or even higher?) may be ok?

AnatolyPopov: But that `0x7fff` is the biggest possible at all, correct? Assuming stack size to be ~1MB then this amount of threads will consume 32GB of memory if they will be allocated simultaneously, right?
In case of 1024 it will be additional gigabyte in the same situation as far as I understand. Are we ok with that?
Also did we do any performance testing with this high parallelism?

jeqo: Of course not :) It's just the upper-bound. The aim of these configs is just to tune, depending on the user's resources -- that's why I'm being relaxed with the upper-bound. 
Next PR is going to add metrics for the thread-pools, so it can be observed if thread-pools are the bottleneck, and users can tune accordingly.

AnatolyPopov: I'm just wondering how tuning this to any higher value affects the performance/memory consumption and if there is any implicit limit. For now this is fine I guess but would be interesting to see how different values will affect benchmarks.

jeqo: Fair point, as soon as metrics start to point into any of these thread pools will start with the perf tuning and share results. There's some discussion here: https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/issues/584#issuecomment-2337889015

AnatolyPopov (approved):
LGTM

PR #597, opened 2024-09-23
feat: bypass chunking when no transformation is requested

jeqo:
Pass directly the file input stream to the storage layer instead of calculating chunks on the fly. Optimizes for the default configuration (no encryption/compression), and prepares for further passing the path to the storage layer if uploading directly from file is preferred.

It contains a couple of meaningful refactoring to clean up RSM logic and simplify chunking. See commits for further details.


core/src/integration-test/java/io/aiven/kafka/tieredstorage/RemoteStorageManagerTest.java
```diff
@@ -479,16 +478,15 @@ void testTransformingIndexes(final boolean encryption) {
             "storage.root", targetDir.toString(),
             "encryption.enabled", Boolean.toString(encryption)
         ));
-        final SegmentEncryptionMetadataV1 encryptionMetadata;
+        final DataKeyAndAAD maybeEncryptionKey;
```
giuseppelillo, line 481: What about using an `Optional` instead of setting `maybeEncryptionKey` to `null` ?
Then when using `maybeEncryptionKey` you don't need to check again if `encryptionEnabled`, but if the value is present in the `Optional` then it means that encryption is enabled and you will use that encryption key. Probably it does not reduce the code but I find it more easy to reason about explicit nulls.

giuseppelillo (approved):

PR #602, opened 2024-10-01
refactor: expose configDef() for all cache config sets

jeqo:
This is needed for documentation purposes (#601), while also improving the cache config builder.

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
@@ -38,59 +38,21 @@ public class CacheConfig extends AbstractConfig {
     private static final String CACHE_FETCH_TIMEOUT_MS_DOC = "When getting an object from the fetch, "
         + "how long to wait before timing out. Defaults to 10 sec.";
 
-    static final long CACHE_RETENTION_MS_DEFAULT = 600_000;
+    public static final long CACHE_RETENTION_MS_DEFAULT = 600_000;
 
-    private static ConfigDef configDef(
+    public CacheConfig(
         final ConfigDef configDef,
-        final Object defaultSize,
-        final long defaultRetentionMs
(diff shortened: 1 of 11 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
jeqo, line 1: I'd start review here: this is the main change affecting the cache config hierarchy. 
Then continue with MemorySegmentIndexes/MemorySegmentManifest configs, and then ChunkCacheConfig and its children

core/src/main/java/io/aiven/kafka/tieredstorage/config/CacheConfig.java
```diff
@@ -38,59 +38,21 @@ public class CacheConfig extends AbstractConfig {
...
-    static final long CACHE_RETENTION_MS_DEFAULT = 600_000;
+    public static final long CACHE_RETENTION_MS_DEFAULT = 600_000;
 
-    private static ConfigDef configDef(
+    public CacheConfig(
         final ConfigDef configDef,
-        final Object defaultSize,
-        final long defaultRetentionMs
+        final Map<String, ?> props
     ) {
```
jeqo, line 46: Disentangle configDef from CacheConfig. Instead the `DefBuilder` defines the config definition.

core/src/test/java/io/aiven/kafka/tieredstorage/config/CacheConfigTest.java
jeqo, line 1: Just adjusting new instantiation.

biggusdonzus (approved):
Code looks good, but i don't understand some of the changes like why in `CacheConfig.java` why move the `configDef.define` defs into the `DefBuilder` class? But maybe that's necessary for the documentation feature, so in the end it's just my ramblings :smiley:

PR #612, opened 2024-10-22
refactor: expose storage backends metric names for doc generation

jeqo:
Add a metric registry for Azure, GCS, S3 metrics, document metrics, and expose metric name templates to be used on documentation generation.


biggusdonzus (approved):
looks good :+1:

PR #615, opened 2024-10-26
MINOR: Explicitly document how the minimum GCS resumable upload chunksize is set

stanislavkozlovski:
I was digging into the project and was trying to figure out how the chunks are set in GCS. After reading this project's code and the GCS SDK's code - I understood how the SDK defaults to resumable uploads. I figured it's useful to explicitly mention how we default to GCS's default when we set it to null and link to the source - for future compatibility, e.g if/when GCS changes their default

### DIFF
--- a/docs/configs.rst
+++ b/docs/configs.rst
@@ -376,7 +376,7 @@ AzureBlobStorageStorageConfig
   * Importance: medium
 
 ``gcs.resumable.upload.chunk.size``
-  The chunk size for resumable upload. Must be a multiple of 256 KiB (256 x 1024 bytes). Larger chunk sizes typically make uploads faster, but requires bigger memory buffers. The recommended minimum is 8 MiB. The default is 15 MiB.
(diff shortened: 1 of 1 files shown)


stanislavkozlovski:
PS: I couldn't explicitly verify, but I'm pretty sure this dictates how many PUT requests the GCS client does internally

stanislavkozlovski:
cc @jlprat @biggusdonzus @jeqo

stanislavkozlovski:
CI failing with:
```
Run if [[ -n $(git status -s) ]]; then
There are uncommitted changes after the task:
 M docs/configs.rst
Error: Process completed with exit code 1.
```
I presume omething in the job itself is broken:
```
  block-fixup:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v3
    - name: Block Fixup Commit Merge
      uses: alexkappa/block-fixup-merge-action@v2
```
Is this the repo for it? https://github.com/alexkappa/block-fixup-merge-action

jlprat:
@stanislavkozlovski yes, it seems the GH action is malfunctioning. We'll take a look this upcoming week.

giuseppelillo (approved):
LGTM

jlprat:
Thanks @stanislavkozlovski !

PR #634, opened 2024-12-06
feat: supports simultaneous upload and download rate limiting

funky-eyes:
The functionality, which originally only limited uploads, has been changed to a global rate limit.

Resolves: 
The purpose of this change is twofold:

When your remote storage service is located in a public cloud, such as AWS S3, and your server is on-premises, pulling data from the remote can easily saturate the local bandwidth, affecting the normal operation of other services.

Generally, AWS EC2's free disk throughput limit is 128MB. When uploading, downloading, and reading from local cache, without a global rate limiter to prevent competition between these operations, disk I/O can become uncontrollable, leading to issues with the producer's write operations.


funky-eyes:
Hi @jeqo，How can I quickly format my code to follow proper coding standards?

PR #650, opened 2025-02-17
fix(storage:s3): limit delete keys to 1000

jeqo:
AWS and other S3 implementtions limit the maximum number of keys to be deleted in a single request[1]
Adding a limit to batch deleteObjects requests.

[1]: https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html


giuseppelillo (approved):
LGTM

PR #660, opened 2025-03-06
fix(build): avoid removing docs on clean

jeqo:
In Intellij, when loading the project and running `clean` phase, the docs are getting cleaned up.
Adding a fix to avoid this.


AnatolyPopov (approved):
LGTM

PR #662, opened 2025-03-10
Support setting storage class for S3 to reduce cost

jiafu1115:
Hi Team:

Thanks for provide so cool tool to support kafka tiered-storage. 

Refer to the document: 
https://aws.amazon.com/s3/storage-classes/?nc1=h_ls
We can know the storage class feature is the core important feature for s3 which involve the cost and performance.
We had adopt tiered storage implement in our Kafka for cost saving goal. thus, we find that all of the Kafka data uploaded by the standard default storage class. So try to create this PR to support the setting for storage class. So that we can get more cost saving.

Thanks in advanced.  WDYT?

FYI:
image
image
image

### DIFF
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
@@ -83,8 +83,8 @@ public class S3MultiPartOutputStream extends OutputStream {
         this.partSize = partSize;
         this.partBuffer = ByteBuffer.allocate(partSize);
         final CreateMultipartUploadRequest initialRequest = CreateMultipartUploadRequest.builder().bucket(bucketName)
-                .storageClass(storageClass)
-                .key(key.value()).build();
+            .storageClass(storageClass)
+            .key(key.value()).build();
         final CreateMultipartUploadResponse initiateResult = client.createMultipartUpload(initialRequest);
         log.debug("Create new multipart upload request: {}", initiateResult.uploadId());
         this.uploadId = initiateResult.uploadId();
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
@@ -37,6 +37,7 @@ import software.amazon.awssdk.services.s3.model.CompletedMultipartUpload;
 import software.amazon.awssdk.services.s3.model.CompletedPart;
 import software.amazon.awssdk.services.s3.model.CreateMultipartUploadRequest;
 import software.amazon.awssdk.services.s3.model.CreateMultipartUploadResponse;
+import software.amazon.awssdk.services.s3.model.StorageClass;
 import software.amazon.awssdk.services.s3.model.UploadPartRequest;
 import software.amazon.awssdk.services.s3.model.UploadPartResponse;
 
@@ -64,17 +65,16 @@ public class S3MultiPartOutputStream extends OutputStream {
     private boolean closed;
     private long processedBytes;
 
-
     public S3MultiPartOutputStream(final String bucketName,
                                    final ObjectKey key,
                                    final int partSize,
                                    final S3Client client){
-        this(bucketName, key, S3StorageConfig.S3_STORAGE_CLASS_DEFAULT, partSize, client);
+        this(bucketName, key, StorageClass.STANDARD, partSize, client);
     }
 
     public S3MultiPartOutputStream(final String bucketName,
                                    final ObjectKey key,
-                                   final String storageClass,
+                                   final StorageClass storageClass,
                                    final int partSize,
                                    final S3Client client) {
         this.bucketName = bucketName;
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
@@ -64,8 +64,17 @@ public class S3MultiPartOutputStream extends OutputStream {
     private boolean closed;
     private long processedBytes;
 
+
     public S3MultiPartOutputStream(final String bucketName,
                                    final ObjectKey key,
+                                   final int partSize,
+                                   final S3Client client){
+        this(bucketName, key, S3StorageConfig.S3_STORAGE_CLASS_DEFAULT, partSize, client);
+    }
+
+    public S3MultiPartOutputStream(final String bucketName,
+                                   final ObjectKey key,
+                                   final String storageClass,
                                    final int partSize,
                                    final S3Client client) {
         this.bucketName = bucketName;
@@ -74,7 +83,8 @@ public class S3MultiPartOutputStream extends OutputStream {
         this.partSize = partSize;
         this.partBuffer = ByteBuffer.allocate(partSize);
         final CreateMultipartUploadRequest initialRequest = CreateMultipartUploadRequest.builder().bucket(bucketName)
-            .key(key.value()).build();
+                .storageClass(storageClass)
+                .key(key.value()).build();
         final CreateMultipartUploadResponse initiateResult = client.createMultipartUpload(initialRequest);
         log.debug("Create new multipart upload request: {}", initiateResult.uploadId());
         this.uploadId = initiateResult.uploadId();
--- a/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
+++ b/storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
@@ -40,6 +40,7 @@ import software.amazon.awssdk.services.s3.model.DeleteObjectsRequest;
 import software.amazon.awssdk.services.s3.model.DeleteObjectsResponse;
 import software.amazon.awssdk.services.s3.model.GetObjectRequest;
 import software.amazon.awssdk.services.s3.model.ObjectIdentifier;
+import software.amazon.awssdk.services.s3.model.StorageClass;
 
 public class S3Storage implements StorageBackend {
     private static final int MAX_DELETE_OBJECTS = 1000;
@@ -47,8 +48,7 @@ public class S3Storage implements StorageBackend {
     S3Client s3Client;
 
     private String bucketName;
-
-    private String storageClass;
+    private StorageClass storageClass;
     private int partSize;
 
     @Override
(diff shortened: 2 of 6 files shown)


jeqo:
@jiafu1115 thanks for this! It looks good, though I wonder if you have considered using S3 Bucket lifecycle to update the storage class on your objects instead of setting them on creation time, and what are the benefits of doing it this way.

Also, if adding new configurations, make sure to update the generated docs (`make docs` would do) to include the changes on your PR.

jiafu1115:
@jeqo thanks for your comments.  
They are different approaches to handle different cases.  
For a lifecycle transition action, It is suitable for the case that we had files which are hot or important in recent days and we apply one storage class. then after some days. it will be less valuable for the older file. so we can using the transition to convert to some other storage class. 
Thus. in my case. we have about 90 Kafka clusters and some are for production business' goal and some are just for log collection or metrics which we already define them we can lost it and prefer to cost saving. So we don't want to configure or maintain too many lifecycle rules to handle it. We just want to set the storage class per cluster according to the cluster's SLA and cost's balance. What's more, the transitions are also charged per request.  we can refer to https://aws.amazon.com/s3/pricing.
So, I think they are different approach for different case which we can combine them to cover all of cases.

WDYT?  thanks for your feedback. If no other obvious issues/concern for this PR, I will update the code according to your suggestion and fix the existed issue such as code style check fail.

Thanks!

jiafu1115:
@jeqo  fix the check style issue and update the document. Can you help to approve pipeline to rerun. thanks

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -64,8 +64,17 @@ public class S3MultiPartOutputStream extends OutputStream {
     private boolean closed;
     private long processedBytes;
 
+
```
jeqo, line 67: nit: new line

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
```diff
@@ -47,13 +47,16 @@ public class S3Storage implements StorageBackend {
     S3Client s3Client;
 
     private String bucketName;
+
```
jeqo, line 50: nit: new line

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -114,6 +120,14 @@ public static ConfigDef configDef() {
                 ConfigDef.NO_DEFAULT_VALUE,
                 ConfigDef.Importance.MEDIUM,
                 S3_REGION_DOC)
+            .define(
+                S3_STORAGE_CLASS_CONFIG,
+                ConfigDef.Type.STRING,
+                S3_STORAGE_CLASS_DEFAULT,
+                ConfigDef.ValidString.in(StorageClass.knownValues()
+                        .stream().map(Object::toString).toArray(String[]::new)),
+                ConfigDef.Importance.MEDIUM,
```
jeqo, line 129: maybe `LOW` as default covers most use-cases?

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -257,6 +271,10 @@ public String bucketName() {
         return getString(S3_BUCKET_NAME_CONFIG);
     }
 
+    public String storageClass() {
```
jeqo, line 274: Let's return `StorageClass` instead, similar to `Region`

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -74,7 +83,8 @@ public S3MultiPartOutputStream(final String bucketName,
         this.partSize = partSize;
         this.partBuffer = ByteBuffer.allocate(partSize);
         final CreateMultipartUploadRequest initialRequest = CreateMultipartUploadRequest.builder().bucket(bucketName)
-            .key(key.value()).build();
+                .storageClass(storageClass)
+                .key(key.value()).build();
```
jeqo, line 87: Should we add a test on S3MultiPartOutputStreamTest to validate that the initialRequest is getting the right storage class via Mockito?

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
```diff
@@ -231,4 +234,28 @@ void withApiCallTimeouts() {
         assertThat(config.apiCallTimeout()).isEqualTo(Duration.ofMillis(5000));
         assertThat(config.apiCallAttemptTimeout()).isEqualTo(Duration.ofMillis(1000));
     }
+
+    @Test
+    void withStorageClass() {
+        final var configs = Map.of(
+                "s3.bucket.name", BUCKET_NAME,
```
jeqo, line 241: nit: change indent to 4 spaces
```suggestion
            "s3.bucket.name", BUCKET_NAME,
```
We should have a better tooling to check this automatically.

jeqo (changes requested):
Made a first review round

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
jiafu1115: ack

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
jiafu1115: ack

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jiafu1115: done. thanks. my bad

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3Storage.java
jiafu1115: done. thanks. my bad

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
jiafu1115: done. I think MEDIUM is also ok due to it is more important when comparing with others. thus, when we consider that maybe some other don't care about it. so I agree to change it to low. thanks

storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfigTest.java
jiafu1115: done. thanks

jiafu1115:
@jeqo  thanks for your code review. Done for the changes.
You can help to double check.  

thanks a lot.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
```diff
@@ -48,6 +49,11 @@ public class S3StorageConfig extends AbstractConfig {
         + "To be used with custom S3-compatible backends (e.g. minio).";
     public static final String S3_REGION_CONFIG = "s3.region";
     private static final String S3_REGION_DOC = "AWS region where S3 bucket is placed";
+
+    public static final String S3_STORAGE_CLASS_CONFIG = "s3.storage.class";
+    private static final String S3_STORAGE_CLASS_DOC = "S3 storage class to store log segments";
```
jeqo, line 54: Actually not only segments but all objects related to it (indexes, manifest).
Maybe a bit more general description:
```suggestion
    private static final String S3_STORAGE_CLASS_DOC = "Defines which storage class to use when uploading objects";
```

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
```diff
@@ -67,14 +68,23 @@ public class S3MultiPartOutputStream extends OutputStream {
...
                                    final S3Client client) {
         this.bucketName = bucketName;
         this.key = key;
         this.client = client;
         this.partSize = partSize;
         this.partBuffer = ByteBuffer.allocate(partSize);
         final CreateMultipartUploadRequest initialRequest = CreateMultipartUploadRequest.builder().bucket(bucketName)
-            .key(key.value()).build();
+                .storageClass(storageClass)
+                .key(key.value()).build();
```
jeqo, line 87: nit: continuation indent seems to be at 8. Maybe change to 4 on this repo?
```suggestion
            .storageClass(storageClass)
            .key(key.value()).build();
```

jeqo (changes requested):
Thanks @jiafu1115! A couple more comments and we should be ready to merge.

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3StorageConfig.java
jeqo: Don't forget to regenerate the docs 👍🏽

storage/s3/src/main/java/io/aiven/kafka/tieredstorage/storage/s3/S3MultiPartOutputStream.java
jiafu1115: you are right. I check the original code. it is right style. So it is my bad. let me correct.

jiafu1115:
@jeqo done with second round's fix. also update the doc. thanks a lot!

jeqo (approved):
LGTM, thanks @jiafu1115 ! 

Lmk if you're up to add a similar feature to the other object storages. Thanks again!

jiafu1115:
@jeqo  thanks a lot ! Let me focus on the S3 adopt in our production in this phase and maybe can take a look at some other storage when free. Anyway. thank you for review. It help us a lot on cost saving.

jiafu1115:
@jeqo  just share my practice for you. maybe you have interesting: When I switch to Standard_IA. the storage cost decrease as expected. Thus. It will cost some extra if one file don't save 30 days. It is aws cost logic. So Standard_IA is suitable for the file storage > 30 days.
Anyway. the storage class is helping us in different case.
image

PR #668, opened 2025-03-30
MINOR: update readme on multipart size

stanislavkozlovski:
https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/commit/96c645d27d57c8f65e896a5203aafc1c1a6cf986 changed the default multipart upload size but didnt update the readme. I figure we can omit the recommendation to increase the value at this point

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,9 +239,8 @@ The plugin rate-limit is complementary to Tiered Storage Quotas.
 When uploading processed segments and indexes, multipart upload is used to put files on the storage back-end (e.g. AWS S3).
 
 > [!IMPORTANT]
-> To control the amount of upload part requests that are executed per segment, each storage back-end has a size configuration.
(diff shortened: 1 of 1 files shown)


README.md
jeqo, line 1: Could have a further change:
- Let's change the section title to "Multi-part uploads" instead of S3... as other SDKs use the same approach internally.
- After mentioning the s3 configuration, let's link to the config page so users can look for more details and other SDKs configs
Sounds good to do this as part of the same PR?

jeqo (commented):
Thanks @stanislavkozlovski! LGTM, but left a comment to improve it a bit further.

README.md
stanislavkozlovski: makes sense. check again @jeqo

jeqo (approved):
Thanks @stanislavkozlovski !

PR #682, opened 2025-05-27
chore(ci): add upload e2e test reports

jeqo:


jeqo:
No need as e2e tests print the output to action logs already

PR #716, opened 2025-08-14
Add Iceberg whitepaper

ivanyu:


iceberg_whitepaper.md
```diff
@@ -0,0 +1,257 @@
...
+- **Coexistence:** Classic (non-Iceberg) and Iceberg segments can coexist in one topic. The segment manifest identifies segment type and file mapping; brokers fetch transparently.
+- **Portability:** Works with REST, JDBC, Hadoop and other catalogs and S3-compatible object storage.
+
+The plugin is responsible for the following:
+- Parse segments as Avro records and validate values against the schema registry.
+- Create the target Iceberg table if absent.
+- Transform Avro values to Iceberg rows and append to Parquet files.
+- On fetch, read the referenced Parquet files, transform back to Avro, and pack records into Kafka batches.
+
+The plugin is not responsible for other operations on Iceberg schemas such as compaction and snapshot expirations. These operations performed by third party tools aren't supported at the moment (see the Upcoming work.)
```
jlprat, line 67: Upcoming work could be a hyperlink to the section down

iceberg_whitepaper.md
```diff
@@ -0,0 +1,257 @@
...
+
+Schema registry is a commonly used component that allows storing schemas and managing their evolution with the desired degree of compatibility. Implementing data lake tables requires schema registry as the storage for Avro schemas and, in the future, as the gatekeeper for schema evolution. There are several open source and proprietary implementations of schema registry. Any of them can be used provided their protocol is compatible with Confluent Schema Registry.
+
+The Iceberg catalog is essential to manage Iceberg tables. It allows managing table schemas, evolving them, and appending data. There are several Iceberg catalog implementations: REST, JDBC, Hadoop, etc. Our solution is catalog-agnostic and supports any of them.
+
+An object storage is required to store the actual data and Iceberg metadata. We implemented the support for AWS S3 (and compatible) as the object storage, however other storages will be added later (see the Upcoming work).
```
jlprat, line 81: Same as above

iceberg_whitepaper.md
```diff
@@ -0,0 +1,257 @@
...
+    C->>-RSM: Commit transaction
+
+    RSM->>+C: Start transaction
+    RSM->>OS: Write segment manifest
+    C->>-RSM: Commit transaction
+```
+
+#### Current limitations
+
+There are circumstances where the same offset may be uploaded multiple times in different segment files. For example, when the partition leader changes with unfortunate timing or an unclean leader election with truncation happens. The write workflow is also not transactional end-to-end. For example, if the Iceberg transaction commits successfully but the remote log metadata persistence fails on the broker side, the broker will eventually re-attempt the uploading. On the next attempt, the data will be uploaded and committed again, thus creating duplicates in the Iceberg table. Kafka brokers have a sophisticated remote log metadata mechanism that allows them to resolve these inconsistencies. Our RSM plugin is compatible with it. However, the plugin doesn't prevent duplicates from being visible to Iceberg readers. It's possible to deduplicate records on the Iceberg reader side. This limitation will be addressed in the future (see the Upcoming work).
```
jlprat, line 164: Same as above

iceberg_whitepaper.md
```diff
@@ -0,0 +1,257 @@
...
+
+Schema evolution is not supported at the moment. The schema of the first record in the topic that the plugin encounters sets the schema of the whole table “forever”. Acknowledging that schema evolution is crucial for most production setups, we're adding it shortly (see the Upcoming work).
```
jlprat, line 166: Same as above

iceberg_whitepaper.md
```diff
@@ -0,0 +1,257 @@
...
+
+Kafka transactions are also not supported for now. If a transaction control batch is encountered, the whole segment cannot be uploaded. The support for transactional data will be added later (see the Upcoming work).
```
jlprat, line 168: Same as above

jlprat (approved):
LGTM only one cosmetic thing that can be left out

PR #722, opened 2025-08-15
Remove Java 11 from pipeline

ivanyu:
As we're starting depending on some broker internals which are built for min Java 17, remove Java 11 from the versions.


AnatolyPopov (approved):
LGTM

PR #724, opened 2025-08-25
Adding support for optional and schemaless keys/values

AnatolyPopov:
Making key and value schemas optional


core/src/main/java/io/aiven/kafka/tieredstorage/iceberg/AvroSchemaRegistryStructureProvider.java
```diff
@@ -66,8 +66,20 @@ public void configure(final Map<String, ?> configs) {
...
+        if (schemaId == null) {
+            return new SchemaAndId(Schema.createUnion(Schema.create(Schema.Type.BYTES),
+                    Schema.create(Schema.Type.NULL)), null);
+        }
+        try {
+            return new SchemaAndId(
+                    Schema.createUnion(Schema.create(Schema.Type.NULL),
+                            (Schema) valueDeserializer.getSchemaRegistryClient().getSchemaById(schemaId).rawSchema()),
+                    schemaId);
+        } catch (final RestClientException e) {
```
ivanyu, line 79: Don't generic network errors fall into this category?

AnatolyPopov: I believe yes but AFAIU it will be thrown only after exhausting all the allowed retry attempts. Or am I missing anything?

Issue #694, opened 2025-06-10
Error Deleting Chunks from Tencent Cloud COS via S3 API: Missing Content-MD5 Header


DDDFiish:
# What can we help you with?

I am using Kafka 3.9.0 in Kraft mode.

When using Tencent Cloud COS as the S3 API backend service, the initial chunking and uploading processes work perfectly, and everything runs as expected. However, when the Kafka message rotates and attempts to delete chunks from COS, an error occurs:

```java
2025-06-10 02:35:56.750291500 [2025-06-10 10:35:56,749] INFO Deleting log segment data for RemoteLogSegmentMetadata{remoteLogSegmentId=RemoteLogSegmentId{topicIdPartition=I7Ell2hpSHi-x3johr37ag:test-1, id=EztBGyCXS5SHaojXz3g_Qw}, startOffset=0, endOffset=10384, brokerId=167844199, maxTimestampMs=1749458025771, eventTimestampMs=1749522955737, segmentLeaderEpochs={0=0}, segmentSizeInBytes=104857345, customMetadata=Optional.empty, state=DELETE_SEGMENT_STARTED} (io.aiven.kafka.tieredstorage.RemoteStorageManager)
2025-06-10 02:35:56.755609500 [2025-06-10 10:35:56,755] WARN [RemoteLogManager=167844199 partition=I7Ell2hpSHi-x3johr37ag:test-1] Current task for topic-partition I7Ell2hpSHi-x3johr37ag:test-1 received error but it will be scheduled (kafka.log.remote.RemoteLogManager$RLMTask)
2025-06-10 02:35:56.755611500 org.apache.kafka.server.log.remote.storage.RemoteStorageException: software.amazon.awssdk.services.s3.model.InvalidRequestException: Missing required header for this request: Content-MD5 (Service: S3, Status Code: 400, Request ID: Njg0NzlhMGNfYmVkMDNkMDlfNWM1ZF81NzA1MWZm) (SDK Attempt Count: 1)
2025-06-10 02:35:56.755627500   at io.aiven.kafka.tieredstorage.RemoteStorageManager.deleteLogSegmentData(RemoteStorageManager.java:689)
2025-06-10 02:35:56.755628500   at org.apache.kafka.server.log.remote.storage.ClassLoaderAwareRemoteStorageManager.lambda$deleteLogSegmentData$6(ClassLoaderAwareRemoteStorageManager.java:94)
[...]

PR #728, opened 2025-08-28
fix(s3): Add LegacyMd5Plugin to S3 client builder

AnatolyPopov:
Resolves: #694

Resolves: #xxxxx
Why this way

### DIFF
--- a/storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3ClientBuilderTest.java
+++ b/storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3ClientBuilderTest.java
@@ -22,7 +22,9 @@ import java.util.Map;
 
 import org.junit.jupiter.api.Test;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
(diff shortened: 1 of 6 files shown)


storage/s3/src/test/java/io/aiven/kafka/tieredstorage/storage/s3/S3ClientBuilderTest.java
```diff
@@ -45,6 +47,9 @@ void minimalConfig() {
             .allSatisfy(metricPublisher -> assertThat(metricPublisher).isInstanceOf(MetricCollector.class));
         assertThat(clientConfiguration.overrideConfiguration().apiCallTimeout()).isEmpty();
         assertThat(clientConfiguration.overrideConfiguration().apiCallAttemptTimeout()).isEmpty();
+        assertThat(clientConfiguration.overrideConfiguration().toBuilder().executionInterceptors())
+            .extracting(ExecutionInterceptor::getClass)
+            .doesNotHaveAnyElementsOfTypes(LegacyMd5ExecutionInterceptor.class);
```
ivanyu, line 52: How this works if by default the config is `false`?

AnatolyPopov: What do you mean? The assertion checks that by default it is disabled.

Issue #731, opened 2025-09-04
iceberg demo fails due to wrong mc usage


hpgrahsl:
# What happened?

When running the iceberg demo the docker compose stack runs `mc` to configure/init minio buckets. This fails with

`mc: <ERROR> `config` is not a recognized command. Get help using `--help` flag.`

# What did you expect to happen?

Should correctly use `mc` to successfully do the minio bucket init. 

# What else do we need to know?

I'll provide the PR to fix this little bug.

PR #732, opened 2025-09-04
fix wrong mc usage for bucket init

hpgrahsl:
Resolves: #731

### DIFF
--- a/demo/iceberg/docker-compose.yml
+++ b/demo/iceberg/docker-compose.yml
@@ -85,7 +85,7 @@ services:
 
   # Minio as S3
   minio:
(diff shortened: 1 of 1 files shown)


AnatolyPopov (dismissed):
Thanks! We need to use fixed minio version to prevent such cases in future. It worked with older image.

hpgrahsl:
@AnatolyPopov I can amend the PR accordingly. Let me know if you have any specific version in mind otherwise I'm happy to pin it to one of the newest ones.

AnatolyPopov (approved):
LGTM

PR #733, opened 2025-09-04
Update configs.rst

tushar1804:

Resolves: #xxxxx
Why this way


ivanyu (changes requested):
This file is generated, please modify `ConfigsDocs.java` as well. You can check if it's working with `./gradlew genConfigsDocs`

PR #763, opened 2025-10-21
Fix closing storage resources

tvainika:
Fix closing storage resources

### DIFF
--- a/core/src/main/java/io/aiven/kafka/tieredstorage/KafkaRemoteStorageManager.java
+++ b/core/src/main/java/io/aiven/kafka/tieredstorage/KafkaRemoteStorageManager.java
@@ -149,7 +149,7 @@ class KafkaRemoteStorageManager extends InternalRemoteStorageManager {
             .findFirst()
             .orElse(null);
 
-        this.segmentManifestCache = new MemorySegmentManifestCache(config.storage(), mapper);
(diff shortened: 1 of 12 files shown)


core/src/main/java/io/aiven/kafka/tieredstorage/KafkaRemoteStorageManager.java
```diff
@@ -148,7 +149,7 @@ class KafkaRemoteStorageManager extends InternalRemoteStorageManager {
             .findFirst()
             .orElse(null);
 
-        this.segmentManifestCache = new MemorySegmentManifestCache(config.storage(), mapper);
+        this.segmentManifestCache = new MemorySegmentManifestCache(storage, mapper);
```
jeqo: I don't recall if there is any purpose on having separate storage instances for fetching manifests. It seems ok to me.

CC @AnatolyPopov in case you know of any special case.

jeqo (approved):

core/src/main/java/io/aiven/kafka/tieredstorage/KafkaRemoteStorageManager.java
avoidik: this is a root cause for broken s3 metrics https://github.com/Aiven-Open/tiered-storage-for-apache-kafka/issues/822

PR #797, opened 2026-02-12
Fix key prefix config name in README

mmaroli:
Updates `rsm.config.storage.key.prefix` to `rsm.config.key.prefix`. 

The key prefix config is on the `RemoteStorageManagerConfig` level so it's not prefixed with the storage prefix. If we use the example config in the README the key prefix isn't picked up and we write to the root path of the bucket.


tvainika (approved):

ivanyu (approved):

jeqo:
Will close and reopen to try to trigger the CI as I can't find the GH action.

jeqo:
Thanks @mmaroli!

